PR #493, opened 2021-06-10
Randomized shopkeepers for Kesmai

jkachhad:
Spawns Shopkeepers that offer randomized wares. There is a chance of spawning a rare item that could not be acquired at such a low level, but at a high price.


surfotato:
I'll take on the jewelers.

Segments/Kesmai.mapproj
```diff
@@ -99075,35 +99096,47 @@ return orc;
...
-        if (Utility.RandomDouble() < 0.2)
-            shopkeeper.Stock<Greatsword>("greatsword", 2000);
-            
-        for (var i = Utility.Random(1, 3); i > 0; i--)
+        /* The vendor won't sell everything though, at least not this time around. */
+        foreach (var stock in wares.Random(6))
+            shopkeeper.Stock(stock);
+
+        /* Of the things they stock, they don't exactly have everything, yet. */
+        for (var i = Utility.Random(3, 5); i > 0; i--)
```
surfotato, line 99139: I don't have a reference for the Utility class, but I thiiiink we want:

```
foreach(StockEntry stock in wares.Random(5))
            shopkeeper.Restock(stock);
```

surfotato (approved):
Was just a minor comment.

Segments/Kesmai.mapproj
```diff
@@ -41,7 +41,16 @@
...
         "Yupo", "Yuqui", "Yura", "Yuri", "Yuta", "Yuti", "Yuto", "Yutu", "Yuxi", "Yuya"
     };
-        
+    
+    public static string[] _rarePurchaseResponses = new string[]
+    {
+        "That's quite a bargain.", 
+        "You won't find another with such quality.",
+        "I didn't think anyone even wanted that.",
+        "It was left here, pray the owner doesn't come back."
```
surfotato, line 50: Missing a comma?

Segments/Kesmai.mapproj
surfotato, line 99139: Since C# has zero-based indices, do you think it'll be clearer for others if we consider 0 the (inclusive) limit wherever we can?

surfotato (commented):
Was just a minor comment.

Segments/Kesmai.mapproj
jkachhad: I did it this way as I think the statement gets re-evaluated when checking the limit of the if statement.  So Utility.Random would generate a new number each time the loop triggered.  We would have to introduce a new local variable, but yes, we should try to make it clearer for others.

PR #521, opened 2021-07-02
Update Axe Glacier.mapproj

SewerCzar:
Added mama and drake spawns, priest recall vendor, MA trainer, ghost should now spawn. Decreased maximum spawns in east ogre cave.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -105631,6 +105631,39 @@
         <block><![CDATA[]]></block>
       </script>
     </entity>
+    <entity name="magicPeacekeeperTemplate">
+      <script name="OnSpawn" enabled="true">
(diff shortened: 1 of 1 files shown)


Segments/Axe Glacier.mapproj
```diff
@@ -105627,6 +105627,39 @@
...
+        
+        Experience = 50,
+    };
+
+    peacekeeper.Attacks = new CreatureAttackCollection
+    {
+        new CreatureBasicAttack(6),
+    };
+    
+    return peacekeeper;
```
jkachhad, line 105656: Peacekeeper doesn't seem to have any spells, not sure if that was just temporary.

jkachhad (approved):
Looks good! Compiled without any syntax issues. Should be up in a few minutes.

PR #526, opened 2021-07-02
Update Axe Glacier.mapproj

SewerCzar:
Added lockpick NPCs. //needs cleanup on NPC templates to reduce total code.

Adjusted Mama ice damage up.

Adjusted Drake attack value down, lightning value up, name changed. He's now immune to his own lightning hopefully.

Martial Artist NPC body type should now be accurate.

Giant chance to hit down, damage up (more accurate to OG I believe)

Attempted to add Lawful gryphon. Logic probably needs work. 

Adjusted spawns down in most zones as it seems I was mistaken on "Maximum" being number of packs, not number of individual monsters. Should be more accurate to OG.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106330,6 +106330,46 @@
       <entry entity="lawfulGryphon" size="1" minimum="1" maximum="1" />
       <location x="35" y="34" region="8" />
     </spawn>
+    <spawn type="LocationSpawner" name="tanner">
+      <minimumDelay>900</minimumDelay>
+      <maximumDelay>900</maximumDelay>
+      <script name="OnAfterSpawn" enabled="false">
+        <block><![CDATA[]]></block>
+        <block><![CDATA[
+
+]]></block>
+        <block><![CDATA[]]></block>
+      </script>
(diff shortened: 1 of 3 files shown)


Segments/Axe Glacier.mapproj
```diff
@@ -104950,9 +104950,9 @@
         new LootPackEntry(true, TestTreasureBottles,    20),
         new LootPackEntry(true, TestTreasureTreasure,   0.63),
             
-            new LootPackEntry(true, (from, container) => new PearDiamond(1500u),    100, gemsPriceMutator),
-            new LootPackEntry(true, (from, container) => new GriffinSkull(),        100, gemsPriceMutator),
-            new LootPackEntry(true, (from, container) => new SilverDagger(),        100, gemsPriceMutator)
+            new LootPackEntry(true, (from, container) => new PearDiamond(3000u),    100, gemsPriceMutator),
+            new LootPackEntry(true, (from, container) => new DrakeScaleArmor(),        100, gemsPriceMutator),
```
jkachhad, line 104954: These should only be available when tanned? May need to remove the drake scale armor.

jkachhad (approved):

PR #575, opened 2021-07-07
Update Leng.mapproj

SewerCzar:
Partial leng spawn, all bosses. Bridge troll is a test monster to see if I can customize damage on mobs with weapons / see if a 100% prone/stun value works.

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -75279,7 +75279,7 @@
         Experience = 1700,
     
     };
-    hobgoblin.AddStatus(new NightVisionStatus(spectre));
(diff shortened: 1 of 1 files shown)


SewerCzar:
Added NPCs with placeholders for king/queen - spawn needs finished on -50 -100 and -200. Unsure of that.

NPC demographics modified but functions remain same in Mausoleum and Gryphon Cliffs (nomads instead of berserkers, vampire familiar style humanoids instead of berserkers / random fighters in vlad temple).

jkachhad (approved):
Looks good, thank you!

PR #615, opened 2021-08-10
Added parameters to spawns

SlysterLOK:
Added parameters so things would actually spawn in the areas I intended. Thanks Sewer!

Also as a test I add a mummy spawn in leng south maus with instant cast curse ( max mana 3 so only 1 cast ). Just seeing if I am getting the hang of stuff or not.


SlysterLOK:
Tweaking the previous

SlysterLOK:
More tweaks. Should be more fluid.

SlysterLOK:
Lowered Kesmai -3 exp a touch.

Another attempt at fixing the dark room south in Kesmai -4. You can see through the top wall area.

PR #663, opened 2021-08-16
Leng Maus - Fix female wizard NPC sounds and add random names for wizard NPCs

RdotSilva:
This PR includes a minor change to the Leng segment. In Leng Maus the female wizard NPCs were using a male chant/death sound. This is a potential fix for that issue. I have also updated Leng Maus NPC names to be randomized similar to the Kesmai segment.

- Fixed female wizard NPC in Leng Maus to use the correct sounds
- Changed wizard NPC names to be randomized similar to Kesmai segment

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -75141,8 +75141,8 @@
                 
     var wizard = new Humanoid()
     {
-        Name = "Thrall",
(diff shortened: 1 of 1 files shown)


jkachhad (commented):
Looks good.

PR #669, opened 2021-08-17
Bloodlands fixes and enhancements

zrikzlok:
fixed some counters and stairs issues also added more spawns..

Fish no longer walk on land!


zrikzlok:
@jkachhad if you happen to have time later today to push it would greatly appreciate ! unless you’re waiting on your entity interaction refactor then I get it :p

PR #689, opened 2021-08-20
Kesmai walls

yarrim:
Trial pass at tile consistency in Kesmai.
Changed the Granite Pillars to indestructible.
Changed floors to statics under indestructible walls and statics to floors under destructible walls.
Added ruins to a few destructible pillars and made portal spikes indestructible.

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -693,6 +693,7 @@
           <wall>31</wall>
           <destroyed>0</destroyed>
           <ruins>0</ruins>
+          <indestructible>true</indestructible>
         </component>
       </tile>
       <tile x="138" y="0">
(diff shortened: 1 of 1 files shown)


jkachhad:
Some of the statics under walls, particularly indestructible should remain static. While converting them to water may make sense, it'll add unnecessary pathing costs.  For example, if the tile is un-pathable, should we really need to worry about drowning in that location? 

This is good way to validate certain things, glad you're working on it! I'll check out the rest later this weekend.

SlysterLOK:
Good job. It is hard to keep track on what needs changed or not. I know some of these I have on my push that has been waiting for a week or so. As long as any overlapping tile changes don't cause issues it should be ok though.

That and people seemingly don't want to provide much feedback on this sort of thing.

PR #704, opened 2021-08-22
World forge

yarrim:
Two commits, one adds a method of caching the required server data, the other adds a new filter and tweaks an existing one.


jkachhad:
I made some changes to the code, let me know your thoughts and we can merge them unless you want to add more.

yarrim:
Those changes look good to me. Your file management is much more robust than mine was. I'll try to match that quality for future pulls.

yarrim:
I'm not getting luck loading components from the cache file anymore*. ScriptingData is getting loaded, but when the startup gets to the ComponentPanel EndInit(), where it parses Core.ComponentsResource, that is coming back null.
ComponentsPanel.xaml.cs:61
*I'm preeeeety sure I successfully used my cached file before, but I had to wait for the server to come back up when I wrote that logic so that I could capture the components in the first place. I may never have successfully used a cached resources file.

yarrim:
Revert revert revert.... but when I backtracked to my implementation, it worked and my 'offline mode' was able to view components.
Makes me think this is an issue related to multithreading, but I've never really developed on anything other than a single thread, so I'm not sure.
If you think you know - I'd rather you give me a pointer than spend your time troubleshooting my code and I'll keep poking at it on my end.

yarrim:
I think this is in good shape now. I plan to do more work, but it will be on a new PR.

PR #767, opened 2021-08-29
Kesmai loot and balance pass test #1

SlysterLOK:
Lowered drop rate of gems but increased their value. Mostly the gem drop chance has been cut in half or slightly more. Gem value has at least doubled or slightly more depending on the dungeon level.

Increased treasure drop chance after -1 as it for whatever reason went from 2.5 in -1 to .6 everywhere else. I have nearly exclusively hunted Kesmai and have found next to nothing as far as treasures and rings. Best bet has been the merchants which is backwards.

Increased potion chance from .1 / .2 to .33 /.35 depending on the dungeon level. This should take the chance from ~1 in 1300+ kills to ~1 in 900+ which is a bit more acceptable.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -103317,7 +103317,7 @@
         new CreatureBasicAttack(6)
     };
     
-    goblin.Wield(new Crossbow());
+    goblin.Wield(new Shortbow());
     goblin.Equip(new LeatherArmor());
             
     goblin.AddGold(220);
@@ -104148,7 +104148,7 @@
         new CreatureBasicAttack(7)
     };
     
-    goblin.Wield(new Crossbow());
+    goblin.Wield(new Shortbow());
     goblin.Equip(new LeatherArmor());
             
     goblin.AddGold(225);
@@ -104448,7 +104448,7 @@
         new CreatureBasicAttack(8),
     };
     
-    fighter.Wield(new FineCrossbow());
+    fighter.Wield(new Crossbow());
     fighter.Equip(new PlatemailArmor());    
     
     fighter.AddGold(550);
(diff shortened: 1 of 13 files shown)


trlook:
I think the reason .6 feels okay in other lands is because of bugged spawn timers and us killing loads more mobs than are intended to be spawned.

SlysterLOK:
> I think the reason .6 feels okay in other lands is because of bugged spawn timers and us killing loads more mobs than are intended to be spawned.

That is possible. Since the Kesmai testing phase began I have found less than 5 'treasures'. And only 1 potion. In Kesmai.

 Granted these items are the main trade-ables but it adds a bit of excitement finding these yourself.

SlysterLOK:
Updated a few mobs to match OG experience values. Some went up and some went down. It was mostly -1 that went up while the other levels had a mix of up and down.

Also tweaked early game crit skill. The surface should more or less be similar to -1 which I think is where most new players will wander to first. Also knocked trog down a bit to be closer to where his lair should be experienced.

jkachhad:
I'll push these changes next week and we can try them out. The exp scaling up might be a big high, but I agree we need to figure out a good gold/gem balance.

SlysterLOK:
> I'll push these changes next week and we can try them out. The exp scaling up might be a big high, but I agree we need to figure out a good gold/gem balance.

If by "we" you mean me because no one else gives a crap about proper balance and flow.

The exp is based off of the archive. There is also a deadzone between -2 and being able to hunt -3 where you need more than -2 offers but there is nothing. In the OG the exp was high in the beginning but hit a wall and stayed there but that wall doesnt exist here. ( Is also why so many areas gave the same exp back then )

SlysterLOK:
Last change for gem test.. Taking a native gem ( in this case the Lair Pear Diamond ) and making it very rare but very valuable in each dungeon level and on each creature.

I started with the chance at 1.5% on Kes -1 with a value of 10,000 and each level increased it be +.3% and +2,000. Just don't know how often these will drop until getting in and testing. May or may not need changing.

This is more or less a look at a possible proof of concept on handling gems in each land. You will have semi rare normal gems that are valuable and 1 native gem that is super rare but very valuable. Kinda like the OG UK Bars but expanded upon.

PR #772, opened 2021-08-30
World forge enhancements

yarrim:
Added scaling and a world render to spawns. Also added a filter to show teleportcomponents and a hotkey to display coordinates on world renders.
Known issue: spawn document renders don't work right until BOTH tabs have had a selection made. Unsure why, I tried cloning the presentation and graphics screens into a region and location specific version with the same issue.


zrikzlok:
Excellent, very excited for scaling and coordinates

PR #795, opened 2021-09-02
Update Underkingdom.mapproj

Wirednuke83:
Added Lightning, WW and PC immunities to TK.


Wirednuke83:
Added the following Changes:

Lurkers immune to Poison
Green Wyverns immune to poison
Orugugra Immune to Poison
Added immunity to magic to Mummy (Councilor), but made him weak to Death and Icespeare
Added Weakness to Death to SwordMaster 
added piercing immunity to skeletons but made them weak to silver

Made Shelob poison weaker, she is currently harder than Carfel.

PR #806, opened 2021-09-04
World forge updates

yarrim:
Added two new 'filters' that don't filter tiles, but highlight respectively:
-Teleporters and destinations. When the selection is on a source or destination tile, the associated destination or source is highlighted in a different color.
-Spawns. All spawns are shown in overlapping rectangles. Useful for finding dead spots or incorrectly ranged spawns.

Added hotkeys to jump directly to Entities (Ctrl-E), Spawns (Ctrl-P), Locations (Ctrl-L), Subregions (Ctrl-U), Treasures (Ctrl-T), and the Segment Document (Ctrl-S). These only work from a world render for right now, but if the selection is over a locationspawner or regionspawner when jumping to Spawns, the appropriate spawn is automatically selected and displayed.

Added Ctrl-Left and Ctrl-Right hotkeys to jump between regions, sorted by RegionID.

Added Backspace hotkey to jump to previous tab. only works on region renders at the moment.

Tweaked Tool and Filter dependencies so that you can now change filters on any tab, since more tabs have world renders. Tools can be changed if the current tab is the ComponentsPanel to streamline world editing.

Final tweak is now there is mousewheel-zooming on world renders.


jkachhad:
I added a `VisibilityOptions` object to control what you're attempting to do with the new filters.  I used to have these options in a very old build and may be good to implement again.  These are just `bool` switches to control how something might render or function.  

Check checks https://github.com/jkachhad/Stormhalter/pull/806/commits/183004d6c5b13ac33c7819b5904c6fdcb1c2e3a5 to see how it can work.

PR #808, opened 2021-09-04
Tile fixes and a few spawn edits

yarrim:
Fixed a few places that got missed in previous pass. statics under secret doors, floors under indestructible walls, etc.
Also changed the Leng bushes to static instead of a tile with two floors. Updated the other floor to have the 2 movement cost instead. this was 30-ish replacements across the regions.
Updated Yasnaki level spawns so that they didn't spawn in Shelob's lair, and had a distinct spawner for the area between Shelob and Carfel.
in Oak, I gave the thief trainer an exclusion region and set the thief portal hexes to affect NPCs in order to keep the wildlife out.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -37972,7 +37972,7 @@
       </tile>
       <tile x="37" y="30">
         <component type="IceComponent">
-          <ground>373020</ground>
+          <ground>20</ground>
         </component>
       </tile>
       <tile x="38" y="30">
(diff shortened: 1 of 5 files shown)


Segments/Axe Glacier.mapproj
```diff
@@ -37972,7 +37972,7 @@
       </tile>
       <tile x="37" y="30">
         <component type="IceComponent">
-          <ground>20</ground>
+          <ground>373020</ground>
```
jkachhad, line 37975: I'm not sure if this was intended. Please let me know and I'll put the PR.

PR #816, opened 2021-09-06
New land (Blackburrow Mines) for testing

Wirednuke83:
Created a new dungeon that will connect to the existing lands in the ogre caves in axe. I made the necessary changes to axe as well, although I didn't know how to link the teleporter. 

I have no populated spawns, but I wanted to be able to walk through it and see what needs to be changed.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -28003,6 +28003,14 @@
       <id>12</id>
       <name>Ogre Caves</name>
       <height>120</height>
+      <tile x="19" y="-5">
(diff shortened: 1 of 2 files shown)


SlysterLOK:
The ogre caves should also have stairs to an Axe UD.. is it in the same general area? Gnolls are awesome enemies / friendlies so I am excited for this one.

Wirednuke83:
Added a town North of the Entrance cavern.

Wirednuke83:
Populated town and corrected a few random issues with tiles.

PR #823, opened 2021-09-07
Kesmai Feedback - September

jkachhad:
Going to go through Kesmai again to check on balance, would like everyone to make a new character and help. The biggest gripes identified are that the difficulty from Kesmai -2 to -3 is huge, particularly for damage output from NPCs.  Secondarily, gold vs gems is under debate, as no one wants to pick up gems.


Wirednuke83:
> Going to go through Kesmai again to check on balance, would like everyone to make a new character and help. The biggest gripes identified are that the difficulty from Kesmai -2 to -3 is huge, particularly for damage output from NPCs. Secondarily, gold vs gems is under debate, as no one wants to pick up gems.

I wouldn't get rid of gems, but if we reduce the quantity of them and raise their value I think it would make finding them exciting again. If you're hunting and a gem randomly drops and it's worth 100k it won't be a hassle to drag it and loot it.

zrikzlok:
have hunted a bit in -2 and -3 now, I think they feel quite good, still chance to die.. (did on my 14 thief) by getting overwhelmed and trying to stick out a fight, but feels quite balanced now

jkachhad:
> 
> 
> > Going to go through Kesmai again to check on balance, would like everyone to make a new character and help. The biggest gripes identified are that the difficulty from Kesmai -2 to -3 is huge, particularly for damage output from NPCs. Secondarily, gold vs gems is under debate, as no one wants to pick up gems.
> 
> I wouldn't get rid of gems, but if we reduce the quantity of them and raise their value I think it would make finding them exciting again. If you're hunting and a gem randomly drops and it's worth 100k it won't be a hassle to drag it and loot it.

I ran a wizard through off the docks, found 4 gems worth 15k in the hour 50 minutes or so I played. I had more gold than I knew what to do with. It gave me a quick start to some jewelry, armor and training. After that, I lost interest in picking up gold or other gems knowing that I could get high-value ones just killing stuff. Especially at 1.5% drop rate. That's 3 gems per 200 mobs.

zrikzlok:
Alright. Played a brand new thaum for a little over an hour tonight mostly in -1 and -2 (ruby) managed to hit lvl 9, only died once and overall had a lot of fun. I think it feels very good, especially for new player experience. 

Stripped my geer and went MA 
Will continue to play more. 

I rarely played a thaum so suggestions below:
Thaum: Perhaps a red box around spells that are likely(will) fizzle since you aren't at the right level. This may be rather confusing / frustrating to brand new players trying to cast all the spells and you can really only cast curse in the beginning. 

I think we need to come up with a solution for magic users and or non fighter/MA weapons starting at the dock. 

It won't be intuitive to strip your armor and weapons and work MA, maybe beef up staff/daggwr a tiny bit to make it semi comprable to MA through black belt/1st dan or so for thieves, wiz and thaum?Also, perhaps class based weapons instead of the typical race assigned ones from OG.  Silly that a thaum or wiz can start with an axe when it really won't be used except for rarer/potential later game cases; definitely confusing for a new player, I suspect they'll assume they should use the weapon they log in with.

Wirednuke83:
> Alright. Played a brand new thaum for a little over an hour tonight mostly in -1 and -2 (ruby) managed to hit lvl 9, only died once and overall had a lot of fun. I think it feels very good, especially for new player experience.
> 
> Rnggods were with me and I got 3 diamonds in about an hour 15 so I was able to keep trained for once, was a nice little treat.
> 
> Stripped my gear and went MA, more on that below.
> Will continue to play more.
> 
> I rarely played a thaum so suggestions below:
> Thaum: Perhaps a red box around spells that are likely(will) fizzle since you aren't at the right level. This may be rather confusing / frustrating to brand new players trying to cast all the spells and you can really only cast curse in the beginning.
> 
> I think we need to come up with a solution for magic users and or non fighter/MA weapons starting at the dock.
> 
> It won't be intuitive to strip your armor and weapons and work MA, maybe beef up staff/dagger a tiny bit to make it semi comprable to MA through black belt/1st dan or so for thieves, wiz and thaum?Also, perhaps class based weapons instead of the typical race assigned ones from OG. Silly that a thaum or wiz can start with an axe when it really won't be used except for rarer/potential later game cases; definitely confusing for a new player, I suspect they'll assume they should use the weapon they log in with.
> 
> Is staff / dagger viable in the beginning? I suppose it's not completely horrible, but compared to MA and it's capabilities in the first few levels vs dock weapons, it doesn't seem close, so seems silly (to experienced players) to use anything but MA... Unless it's changed and I'm just stuck in the old ways.. If so please let me know so I can start working staff since I just made black belt :p.
> 
> Also new player experience related. We need a starter scroll (perhaps class based) or something that tells players you can ask trainers name, train, teach, show prices, sell, etc. Even being experienced I forget sometimes.
> 
> Perhaps class based scrolls with some general hints to the trainers etc. A brand new thief probably will have no idea about secret doors or how to find the thieves guild.

There is a lot of room to expand into other weapon types. We could make caster daggers and wands... really what is stopping us? If OL staff/Hand becomes the META again... I think we've failed.

SlysterLOK:
> > > Going to go through Kesmai again to check on balance, would like everyone to make a new character and help. The biggest gripes identified are that the difficulty from Kesmai -2 to -3 is huge, particularly for damage output from NPCs. Secondarily, gold vs gems is under debate, as no one wants to pick up gems.
> > 
> > 
> > I wouldn't get rid of gems, but if we reduce the quantity of them and raise their value I think it would make finding them exciting again. If you're hunting and a gem randomly drops and it's worth 100k it won't be a hassle to drag it and loot it.
> 
> I ran a wizard through off the docks, found 4 gems worth 15k in the hour 50 minutes or so I played. I had more gold than I knew what to do with. It gave me a quick start to some jewelry, armor and training. After that, I lost interest in picking up gold or other gems knowing that I could get high-value ones just killing stuff. Especially at 1.5% drop rate. That's 3 gems per 200 mobs.

I need to see a replay. None of that syncs up with what anyone else has told me. You will fall behind on training fast if you just wait and pick up the high value rare gems. And the 1.5% rate is for the 10K gems on -1. Not 15K. You arent doing much of anything only picking up 3 rare gems every 200 kills either way. The rate may need to come down like I DMed you since at that time it was a blind guess as to what would feel like a nice surprise drop.

However beyond the gems. The gold drops on mobs is clearly set too high now. Even with the increased value gems players are walking out with as much coin as they are in gem value. Gold needs be be dropped to half or more.

zrikzlok:
After hunting in -4 an hour or so today I’ll agree that gold drops are a bit high. Outside of that though, difficulty feels pretty good.

PR #868, opened 2021-09-13
World forge improvements

yarrim:
* Spawns Document now initializes its renderers, so it no longer seems broken at startup.
* Spawn and Teleport overlays use new VisibilityOptions to show\hide.
* Added Show Secret Doors visibility option to show secret doors as their staticID when enabled.
* Components Window hides when the draw or paint tools are not selected.
* When placing a floor, water, ice or air component, the existing components of all four types will be removed. Hold shift to append, if desired.
* Pressing Ctrl-D when the current selection is a teleporter (stair, air hex, hidden teleporter, etc) will change the current tab to the destination region, select the destination tile, and attempt to center the screen on it.
* Added a few lines of calculated statistics to RegionSpawners: Average Mobs, Open Floor Tiles and Density.
* Ctrl- Hotkeys to jump to documents: s**P**awns, **E**ntities, **T**reasures, **L**ocations, s**U**bregions, se**G**ment. Ctrl + Left and Right arrow keys also cycles through regions in order of their IDs. 
  * When using these hotkeys, if an appropriate tile is selected, the destination document will automatically display the related entry. I.e. jumping to the spawns document while having the selection on a location spawner will jump to that location spawner's entry.
* Ctrl-S now saves the segment.


yarrim:
Feel free to cherry-pick these commits if any are problematic.

'Migrate Spawn and Teleport filters to VisibilityOptions'
  should include 'Show Secret Doors visibility option' - but the latter is just an icon.

'Show\Hide Components window when DrawTool is selected'
 should include 'Improved Show\Hide of ComponentsWindow'

I think the rest are independent. I tried to commit everything as complete changes.

If you only have a few minutes, I think 'SpawnsDocument fixes' is the most valuable, since it removes a pain point anyone working on spawns is currently feeling and is two very small changes: calling the initialize() method of the presenters and changing how regions spawns are 'centered'.

jkachhad:
Great changes. I'll merge this in tonight and push it out with the update.

PR #870, opened 2021-09-13
Minor Fix to Oak Salamanders

Wirednuke83:
Changed the following:

- Made salamanders immune to fire
- Turned off their ability to loot since they are monster/creatures.


jkachhad:
Converted Salamanders to inherit from `AnimalEntity` which do not loot by default.

PR #909, opened 2021-09-20
Oakvael - Garg Weakness Update

Leolus:
- Added Silver weakness to gargoyle to match previous iterations.


jkachhad:
We may need to work on consistency across the type, rather than per land/entity?  Should all skeletons be immune to piercing, as gargoyles are weak to silver.  This would help later on when designing zones or encounters, if a player encounters a skeleton, they know their rapier will not work, and that could be a life saving decision.  Rather than guessing if their rapier will work on a particular spawn of a skeleton.

PR #910, opened 2021-09-20
UK - Princess Set to Passive and added trail

Leolus:
- Lowered UK Princess' desire for blood and gore.
- Added trail option to princess.


Chilly021:
I tested this shortly after reset and also again yesterday. It looks like Princess is still aggressive. While following, she will run to crits and even chase them until they are dead. This is really a pain on archers.

PR #1004, opened 2021-10-28
Miscelaneous/Book/Ring Base

SewerCzar:
Added baselines for the unused ring models, books, and a few miscellaneous items.


SewerCzar:
Fixed and transferring to new pull

PR #1013, opened 2021-10-30
Axe Undead updates

SewerCzar:
Pushed experience changes for axe undead to even out with grove experience values.


SewerCzar:
Added a youth potion drop change. Adjusted dog numbers down. Treasure chance was .1 instead of 1. Popped it up to .63% like other monsters.

PR #1040, opened 2021-11-07
World forge

yarrim:
Enhancements:
* Paint tool now behaves like Draw tool in that a painted floor-type component will overwrite other floor-type components. Append instead with Shift. Paint ONLY the selected component and clobber all existing components with ALT.
* When using the delete hotkey, Only currently visible components are deleted from the selected tile. Filter to floors to remove floor-type components, etc.
* When pasting from the clipboard, the behavior is changed if the target selection is more than 1x1. Old behavior, or if the target selection is 1x1: whole clipboard is pasted in starting with the top left tile as the current selection. New behavior if the target selection is larger than 1x1: the clipboard is tiled horizontally and vertically to fill the selected space. Clipboard tiles that extend beyond the selection are not applied.
* Tile Component Editor (double click a tile) now has a delete action button to remove a specific component without removing all other components.
* Counter property editor now reads and writes accessDirection. This is a bit of a workaround. Only the LAST entry selected will be saved to the mapproj. The editor does not show that previous values are cleared unless you close the component editor or navigate to another component on the tile and back.
* When editing components, changes to color, static IDs and other properties are periodically rendered on the world screen. Previously, the editor had to be closed and the world moved to trigger a new render.
* Using the Ctrl-T hotkey to jump to the Treasures tab will now jump directly to a Treasure definition if coming from the Entities tab if the currently selected text matches a Treasure name.
* When jumping to Entities, Spawns or Treasures, the list will scroll to bring the destination entry into view.
* Treasures Document has a listing showing all Entities using that Treasure name. This is not code interpretation, but simple string matching. An Entity with a comment containing the Treasure name will still appear in the list.
* On the Entities, Spawns and Treasures documents, the lists of related objects have a right-click context menu that will jump directly to that record in the related tab.

Fixes:
* Lists for Entities, Spawns, Locations, Subregions, Treasures all now correctly receive mouse scroll input
* A few sprites (ones that were not 100x100 and had offsets, will now render correctly when zooming in and out. Thanks Midgit.

Entity, Spawn, Treasure, and Location hotkeys only jump to a specific entry if the tab has already been loaded. The Documents don't initialize and register themselves for the event I send until the user opens them manually once. If anyone knows how to force them to initialize on segment load, please let me know. My Google-fu has failed me.

### DIFF
--- a/Source/Kesmai.WorldForge/Editor/Clipboard/ClipboardManager.cs
+++ b/Source/Kesmai.WorldForge/Editor/Clipboard/ClipboardManager.cs
@@ -90,38 +90,108 @@ namespace Kesmai.WorldForge.Editor
 			var presenter = ServiceLocator.Current.GetInstance<ApplicationPresenter>();
 			var document = presenter.ActiveDocument;
 
-			if (document is SegmentRegion region)
+			if (document is SegmentRegion region && selection.Region.ID==region.ID)
 			{
 				var areas = selection;
 				var area = areas.FirstOrDefault<XNA.Rectangle>();
 
 				if (area != default(XNA.Rectangle))
 				{
-					if (Clipboard.GetDataObject() is DataObject data && data.GetDataPresent("MemoryStream")
-					    && data.GetData("MemoryStream") is MemoryStream stream)
+					if (area is {Width:1,Height:1})
+                    {
+						PasteToOne(area, region);
+                    } 
+					else
+                    {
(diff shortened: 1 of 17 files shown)


yarrim:
RE: accessDirection: @jkachhad I'm not sure how to get this to behave more like an enum than a comboSelectBox. Feel free to reject this commit (#ea985b2) if you think this should be done in another way.

PR #1047, opened 2021-11-08
Kesmai hol teleporter

Leolus:
- Moved the drop one hex west to avoid teleporting back out.


Leolus:
Fixed by another pull

PR #1051, opened 2021-11-10
Kesmai - Hall of Legends surface stairs fix

Tawnos-LOK:
The surface stairs to the Hall of Legends are currently broken.  I looked at the hex and it appears the wrong region was utilized.  This fix sets the stairs hex on the surface to match the stairs hex in the Hall of Legends region [9].


jkachhad:
Originally there was a set of stairs hidden in the locker room that gave you access.

zrikzlok (approved):

PR #1056, opened 2021-11-11
Quest items bound to killer

SewerCzar:
Crystal Gaunts and Knight Halberd will now check and (hopefully) bind to killer if they are of the right class.


SewerCzar:
CLOSED due to cg quest

PR #1085, opened 2021-11-17
Axe Lockpick Thief Trainer Addition

Leolus:
- Added Thief trainer to lockpick town


jkachhad:
Added the thief trainer to the initial entrance rather than in Lockpick town.

Leolus:
The purpose of this was to allow a high level magic trainer for thieves that you could push gold to. Other classes can do this but due to teleporter placements it's not possible for a thief to push gold to train. This could also be mitigated with some kind of gold carry solution but right now training higher level magic on a thief is very painful.

PR #1087, opened 2021-11-17
Bloodlands morbin damage fix

Leolus:
- Spacing update


Leolus:
@zrikzlok Have any issue with this?

zrikzlok:
I had set it at 1 purposely, hadn’t heard any complaints so far /shrug other than the shock factor,  if there are complaints and  or community wants it.. or can set it to 50 so it’s not so drastic ?

Leolus:
Okay I'll close then, I was under the impression it wasn't intentional. The only issue I can see with it is people fighting near the town entry getting knocked to 1 then killed by a mob.

PR #1089, opened 2021-11-17
UK - Fixing Carful walls

Leolus:
- Added indestructible flag to additional walls to prevent non-thief entry


SewerCzar:
I have a wall fix in my "pending UK issues" pull already. This is probably a conflict but not sure

Leolus:
I'll close this due to fix being implemented in Sewer's fix.

PR #1136, opened 2021-11-28
Spawn for thief trainer in Axe

yarrim:
Created spawn and used existing (but previously unused) entity.


yarrim:
Might conflict with #1085, but we placed him in different locations. My change adds a trainer to the thief-only cave under Lockpick town. Leo's adds one to the pub.
I never used Axe for training in OG, so I don't remember where the trainers should be.

PR #1144, opened 2021-12-01
World forge

yarrim:
Enhancements:

*Added a Help Menu with a ""What's New"" window and a link to the WorldForge article on the wiki.
*Added a calculated 'Threat' metric to Entities. This (m,r) notation shows skill levels in melee and ranged damage respectively. Magic skill levels are doubled (that coefficient may need tweaking). This threat metric, as well as HP and XP are now presented in the datagrid for Spawners.
*Component Editor has two new buttons for reordering components. This generally has no effect on gameplay, but certain components can layer.
*Component Editor supports Ctrl-C and Ctrl-V for copying and pasting individual components. This and other Copy\Paste functions below use the clipboard to store XML.
*Spawners, Entities, Subregions, Locations and Treasures all have export, clone, and import buttons where the 'add' and 'delete' buttons are.
*Improved right-click context menu on the world renders. Selections can be right clicked and have menu options to turn them into spawners or subregions, or to add the selection as an inclusion or exclusion to a regions spawner.
*Added a button to the Component Editor that allows quick selection of the destination tile for TeleportComponents like stairs, ropes, and hidden teleporters. Click the 'Select Destination button, then right-click the target tile and 'Set as teleporter destination..'

Fixes:

*Fixed crash related to having an old segment's spawn or entity document still around and responding to messages after changing to another segment.
*Alert, instead of crash, when trying to load a segment from a file that isn't a correctly formatted mapproj.


yarrim:
Example of the teleport destination selecting.
targetTeleporter

yarrim:
Example of quick spawner editing.
spawnerEditing

zrikzlok:
awesome additions, thank you very much

PR #1151, opened 2021-12-04
Search for experience assignment from syntax tree.

jkachhad:
For @yarrim as a concept on how to search through a SyntaxTree of a particular script block. This might catch issues where experience is assigned twice, or in a way that the regex fails.


yarrim:
This was an example for me, and implemented in 468ba70. This PR can be closed.

Issue #725, opened 2021-08-24
weak str rings bugged, give +3 str


stormhalter:
also weird behavior with how putting on equipmenet adjust stats. [Reported by HealthySnack]

trlook:
> 
> 
> also weird behavior with how putting on equipmenet adjust stats. [Reported by HealthySnack]

Okay, so from further testing, no damage bonuses up to 18 character sheet strength.  Adding +6 strength also gives no bonus damage.  After that, get bonus damage for each +3 up until 36 strength:

- 18 str (no dmg bonus)
- 24 str (no dmg bonus)
- 27 str (dmg bonus)
- 30 str (dmg bonus)
- 33 str (dmg bonus)
- 36 str (dmg bonus) (cap)

For similar behavior to dex, a damage bonus should be awarded at both 18 and 24 strength. However, not sure if this is the intended behavior or not.

PR #1223, opened 2022-01-27
Fix weak strength bonus from 3->1 (Address issue #725)

trlook:
Fix weak strength bonus from 3->1


jkachhad (approved):
Interesting this went unnoticed. Thanks!

Leolus (approved):

Tawnos-LOK (approved):

PR #1228, opened 2022-01-27
Kwolf Randomized Spawn

SewerCzar:


zrikzlok (approved):
Randomization to spawn and slight increase to difficulty look reasonable.

Chilly021 (approved):

Chilly021:
Is it possible to have more than one kwolf up with this code?

SewerCzar:
Chilly: Theoretically no. The spawn will change locations not die, which should cause no new spawn thread on teleporting. Spawner doesnt search for monster in range, it only triggers on death.

Leolus (approved):

PR #1229, opened 2022-01-27
bloodlands fix for flordex

zrikzlok:
fix for flordex... maybe?


Chilly021 (approved):

Segments/Bloodlands.mapproj
```diff
@@ -96,7 +96,7 @@
 
         public override void OnSpellTarget(Target target, MobileEntity combatant)
         {
-            if (target != null)
+            if (target = null)
```
Leolus, line 99: Should this be == not =? Single ='s are normally used to assign a value where as == are equality checks.

zrikzlok: Ugh yes I think you’re correct needs ==

zrikzlok: Are you able to make that correction ? Won’t be at a computer for another few hours

Leolus: It won't allow editing on github because of the file size and I'd have to open a completely new pull for the change. Probably best to wait until you can update it.

zrikzlok:
alright, pushed the fix for ==

Chilly021 (approved):

SewerCzar (approved):

Leolus (approved):

PR #1230, opened 2022-01-28
Add Jumpkick instead of Charge

Chilly021:


Chilly021:
Let's go with Sewer's implementation.

PR #1232, opened 2022-01-28
Updates and Fixes for Blackburrow

Wirednuke:
- Fixed the portable hexes that allowed you to portal around Jym the Giant
- Fixed the teleport hex in the thief bypass to allow mobs to chase you into the secret room. 
- Removed the region spawn from the stairs down to -1 so you don't get inundated with mobs the second you come downstairs. I think this will need more balancing but I want to make small adjustments.
- Added a few roaming skeleton archers with high hide detection to make the final encounter more challenging for thieves. 
- Lowered the drop rate slightly on the high end, high value gems.


Tawnos-LOK (approved):
Code looks good to me and is consistent with comments.

zrikzlok (approved):
Looks good to me

Chilly021 (approved):

PR #1253, opened 2022-02-02
Big ass leng overhaul

SewerCzar:
Changed all monsters to templates (surprisingly small balance changes - Vlad has slightly more damage and vastly less hps, Kosh has a lot less experience)

Nearly all monsters have higher base dodge, but the majority of it should be overcome by weapons / small amounts of skil. 

Some casters might be more dangerous.

Leng Undead is populated, and dangerous.

Gems that will change are done. Gotta find that sweet spot.

Experience has changed a bit, but is a lot more consistent.

Wands and orbs are more consistent

Bottles are more consistent

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -73,6 +73,99 @@
         
         item.Price = Utility.RandomBetween(-lesserPrice, higherPrice);
     }
+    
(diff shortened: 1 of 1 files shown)


SewerCzar:
Please to make sure no one edit map before make edit

PR #1254, opened 2022-02-02
Update Leng.mapproj

SewerCzar:
Changed all monsters to templates (surprisingly small balance changes - Vlad has slightly more damage and vastly less hps, Kosh has a lot less experience)

Nearly all monsters have higher base dodge, but the majority of it should be overcome by weapons / small amounts of skil. 

Some casters might be more dangerous.

Leng Undead is populated, and dangerous.

Gems that will change are done. Gotta find that sweet spot.

Experience has changed a bit, but is a lot more consistent.

Wands and orbs are more consistent

Bottles are more consistent

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -73,6 +73,99 @@
         
         item.Price = Utility.RandomBetween(-lesserPrice, higherPrice);
     }
+    
+        public static void gemsPriceMutatorLow(MobileEntity entity, Container source, ItemEntity item)
+    {
+        /* Item values fall between 0.8 to 1.2 of base price. */
+        var lesserPrice = (int)(item.BasePrice * 0.6);
+        var higherPrice = (int)(item.BasePrice * 0.2);
+        
+        item.Price = Utility.RandomBetween(-lesserPrice, -higherPrice);
+    }
+    
+    public static void gemsPriceMutatorMedium(MobileEntity entity, Container source, ItemEntity item)
+    {
+        /* Item values fall between 0.8 to 1.2 of base price. */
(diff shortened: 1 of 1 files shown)


zrikzlok (approved):

Segments/Leng.mapproj
```diff
@@ -79351,7 +79822,7 @@
 ]]></block>
         <block><![CDATA[]]></block>
       </script>
-      <entry entity="cliffBossKosh" size="1" minimum="1" maximum="1" />
+      <entry entity="BossLevel11Kosh" size="1" minimum="1" maximum="1" />
```
Leolus: Boss levels seem a little odd. Power Troll is definitely weaker than Kosh but is rated at level 12 to Kosh's 11? Power Troll is for NV before Kosh and key for princess quest key for the stat pot.

SewerCzar: He's a miniboss. I can reflect the power level differences in a future pull - Some monsters are set to the highest end variable, some have a multiplier of 3 on hit points, +3 attack +3 defense from their basic model, and some have 6x hps and a few other variables - see Drake/Vlad. They're on that format now (Drake didn't change at all essentially. Kosh lost hps)

Leolus:
I think the naming conventions will make it harder to find and adjust specific monsters. Instead of just looking for Maus items for blanket level changes you have to remember/find out which of the level monsters are each level. It's not a big deal but if we have different standards on different lands it's going to get nasty trying to track what to change.

Leolus (approved):

Chilly021 (approved):

PR #1263, opened 2022-02-04
Update Underkingdom.mapproj

SewerCzar:
Monsters normalized - mostly decreases to hit points. 

Carfel is harder to hit but less hit points

Undead stalkers no longer cast death

Gems are rarer and worth more on average

Troll King is no longer immune to death

Spawns have been slightly spread out so the entire zone doesn't get locked behind one secret door.

The wyvern complex is no longer indestructible on the surface.

Monsters all have slightly fewer hit points but higher defense. Shouldn't affect many players but blocks will probably occur on normal mobs, and almost certainly will on high tier monsters. Their hit points have almost all been decreased though.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -102272,7 +102272,7 @@ CreatureEntity OnSpawn()
     var hobgoblin = new Hobgoblin()
     {
         MaxHealth = 120, Health = 120,
-        BaseDodge = 14,
+        BaseDodge = 20,
         HideDetection = 25,
         Experience = 5015,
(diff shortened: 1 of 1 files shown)


zrikzlok (dismissed):

Chilly021 (dismissed):

Leolus (dismissed):

yarrim:
hold off on comitting this. There are some syntax errors. I'll submit fixes shortly

yarrim (approved):

Leolus (approved):

Chilly021 (approved):

PR #1299, opened 2022-02-07
World forge 2 2

yarrim:
Enhancements:
* Threat calculation uses Code Analysis instead of regex to determine skill levels.
* Added a visibility option to show walls and doors as their destroyed images.
* Added a Flags field to Entities. Visible on the Entity tab as well as the grid view on a Spawner. Shows statuses and attributes like canFly, canLoot, NightVisionStatus, Poison and Prone. Can only show values if they are explicitly specified in the entity OnSpawn() script.


SewerCzar (approved):

zrikzlok (approved):
I are excited!

Chilly021 (approved):

PR #1302, opened 2022-02-08
Update Oakvael.mapproj

SewerCzar:
Stat squish of oakvael like Kesmai -

Overall not much changed. Wendy is now a "minor dragon" but her stats should be good. They're like the revisit but...

She's easier than drake except the whirlwind 

Lich's slightly different, most monster spells are down, several outliers are up.

Experience overall is up.

Biggest changes are basedodge which I think has been successful so far as baseline was comically low before


zrikzlok (approved):
Will be good to see how this goes, seems reasonable and kesmai seems good so far on difficulty scale

Leolus (approved):

Chilly021 (approved):

PR #1331, opened 2022-02-12
UK - UD Dex Ring

Leolus:
* Added +2 Dex at 2 weight to UD treasure. This works out to 3.70% and is the lowest drop rate available there.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -105540,6 +105540,15 @@
           <block><![CDATA[]]></block>
         </script>
       </entry>
+      <entry weight="2">
+        <script name="OnCreate" enabled="true">
+          <block><![CDATA[]]></block>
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

SewerCzar:
Disagree with this. Plenty of reasons to go to UK UD already. Item isn't really appropriate here.

Leolus:
> Disagree with this. Plenty of reasons to go to UK UD already. Item isn't really appropriate here.

Item isn't obtainable outside of Leng Drake at all at a 5% drop rate. Where do you expect people to farm this?

zrikzlok (approved):
Seems like a very tiny drop rate chance. Willing to give it a shot and see how it looks with someone farming down there for a few hours.

Tawnos-LOK (approved):

PR #1340, opened 2022-02-14
Zoo idea test portal

Wirednuke:
An idea to help with zoo collection in secret rooms:

A teleporter hex that only allows chaotic mobs, so when they wander onto it they get moved out of the zoo room. 

This is just one test hex on a secret room on -1 that collects alot of mobs quickly and effects population density on the dungeon level.


jkachhad:
Rather than implementing this, I think we should change the internal spawners to calculate densities and spawn farthest away from high density.

PR #1364, opened 2022-02-24
Frost giant minor update

SewerCzar:
Gems will now be Undead area gems at 10x drop rates from other monsters

Frost giants should use concussion as originally intended now


Leolus (approved):

Chilly021 (approved):

Tawnos-LOK:
The walls by Vile Necro are not indestructible so the concs will end up freeing him down there.

zrikzlok (approved):

PR #1367, opened 2022-02-25
Grove Fixes

SewerCzar:
Removed artifact segment code

Grove Heart monsters will now all be evil

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Oakvael" version="0.90.0.0">
+<segment name="Oakvael" version="0.92.0.0">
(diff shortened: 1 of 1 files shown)


Leolus (approved):

Chilly021:
Why evil? Just didn't know. The reason I would say no to this is because people could easily take the defenders into the real area and kill all other crits.

SewerCzar:
> Why evil? Just didn't know. The reason I would say no to this is because people could easily take the defenders into the real area and kill all other crits.

All the monsters down there were evil already prior to my changes right before live. As things stand they're currently fighting each other in an eternal war. It needs to be changed back.

This is a reversion to the old way, not a new thing. And the nature of the spawns prevents a big war breaking out.

zrikzlok (approved):

Chilly021 (approved):

PR #1368, opened 2022-02-25
OL Staff Rebalance

Tawnos-LOK:
Adjusted OL staff defensive stats to be more in balance with class scaling:  
+1 tick additional BaseArmorBonus (bringing the total to 5)
+3 Projectile Protection added to be in line with the new wand

Prior to this change, a MU with 24 combat adds, +2 Dex, Super Icies, and 18.5 skill only has 1730 DR whereas an 18 MA with 9th Dan (15 Skill), CGs, and +2 Dex already reaches this level.


Chilly021 (approved):

Leolus:
I'm okay with this but we might want to consider talking to Lux about the base DR on staff weapons. If we have to give staffs a lot of DR to make them viable this might be better done on the base class.

Leolus (approved):

starfire5:
It would be great if this could get full approval as the Overlord staff is one of the more essential weapons for a caster.

PR #1370, opened 2022-02-26
shield change proposal

Wirednuke:
Proposal to make shields a little better and more useful:

- Set wooden shield to +1 projectile protection
- Set Plated shield to +2 projectile protection
- Set steel shield to +3 projectile protection
- Set lightning shield armor bonus to 2, and +4 projectile protection.

These changes would help everyone using shields.. and would pair well with the new caster wand to give casters more survivability. 

They would provide a boost to the mace/shield combo for knights, but wouldn't over power any of the other weapons.

### DIFF
--- a/Source/Kesmai.Server/Game/Items/Equipment/Shields/LightningShield.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Shields/LightningShield.cs
@@ -19,10 +19,10 @@ namespace Kesmai.Server.Items
 		public override int Category => 1;
 		
 		/// <inheritdoc />
-		public override int BaseArmorBonus => 1;
+		public override int BaseArmorBonus => 2;
 
 		/// <inheritdoc />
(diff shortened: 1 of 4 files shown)


Chilly021:
While I would selfishly love this I think we are slowing down on base changes. I'll put lux on here.

Wirednuke:
> While I would selfishly love this I think we are slowing down on base changes. I'll put lux on here.

I know, just figured if I did all the work y’all could just say yes or no. I think having more progression on the shields makes sense, and it’s not a game breaking change.

SewerCzar (commented):
I agree with Chilly. This is a Lux call for how he decides to deal with Archers in the long term

Wirednuke:
> I agree with Chilly. This is a Lux call for how he decides to deal with Archers in the long term

All good, just easier for me to just write and show you my idea rather than explain them. Not that great with the voice chat verbal communication business lol.

Issue #739, opened 2021-08-25
halberds trigger extra defenses on target


stormhalter:
attacking with halberd causes the defender to gain double the basearmorbonus of their weilded items, perhaps this is because it has both slashing and bashing properties? [Reported by HealthySnack]

trlook:
Needs to be fixed in the Armor.cs class. The implementation of the fix requires design choices.

sproatba:
@Chilly021 I think this can be closed

PR #1388, opened 2022-03-05
WeaponFlagFix

trlook:
Should fix weapons with multiple flags seeing stacked armor. Highly recommend this gets tested on a local build first. Addresses issue  #739 halberds trigger extra defenses on target

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.cs
@@ -1,0 +1,30 @@
+namespace Kesmai.Server.Game
+{
+	public abstract partial class MobileEntity : WorldEntity
+	{
+		/// <summary>
+		/// Calculates the damage modifier against the specified item.
+		/// </summary>
+		public int CalculateDamageModifier(ItemEntity item)
+		{
+			var paperdoll = Paperdoll;
+			
+			if (paperdoll is null)
+				return 0;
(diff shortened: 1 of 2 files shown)


SewerCzar (dismissed):
I like the change but I'm not qualified to check that code. Looks good on my end

dedmist (approved):
I'm assuming that ArmorProtections is defined to determine which single bonus to use.  Looks fine.

jkachhad:
What happens in the case that an armor protects against multiple damage types, and against a weapon that deals multiple damage types?  Does it just use the minimum?

trlook:
> What happens in the case that an armor protects against multiple damage types, and against a weapon that deals multiple damage types? Does it just use the minimum?

The original implementation was that a multiple type weapon like halberd (blunt/slash) would see both types of defense. Almost all armor has blunt/slash/pierce protection in equal measure, so it would get double penalized before. Now, it should just check which of the protections is lowest as if smartly attacking with the weapon to be more effective.  We could also average the protections instead such that some of the attacks are more effective if the defender is lacking a protection against a type.  But since most things have multiple protections anyways, the usual behavior will be that a multitype weapon will perform the same as a single type weapon instead of worse.

jkachhad:
Looks good. I changed list to just a byte array to minimize memory allocations since this will likely execute quite often.

jkachhad (approved):

Leolus (approved):

PR #1393, opened 2022-03-06
DamageMitigation

trlook:
Allow for armorBonus to mitigate damage right away.

This will provide a lot of mitigation vs packs of monsters (and archers!).


Leolus (dismissed):

dedmist (dismissed):
Looks good, discussed this in chat already.

PR #1395, opened 2022-03-06
Attempt to implement item breaking on death and random prop items

trlook:
An attempt to implement items breaking upon death due to durability loss via IFragile interface. Test item is a random property spear. Note, I'm not sure if I can just override with a random number, or if the randomization needs to be done in the instantiation so I've implemented half the properties each way to test.


sproatba:
item durability has historically been covered by charges.  Is there a need to add this in addition to charges?

trlook:
This would be for items intended to break (i.e. be deleted at depletion).  I could see uses for item being both charged and Fragile.

Leolus (approved):
Just a note that you won't be able to use the new item until Lux recompiles the client unless you put the item etc in a segment.

trlook:
I'm going to close this for now as it needs some work for the actual random item properties to serialize, so shelving this for now

PR #1397, opened 2022-03-07
Modifiable Items

trlook:
Implement a Modifiable Leather Armor as a proof of concept for modifiable items.  Implement Rarity static class to assist with tinting and labeling modified items. The leather armor should take an int for the amount of BaseArmorBonus. This works for armors because most armors implement defenses as protections isntead of base armor. For Weapons, we would want the argument to add to the existing BaseAttackBonus and BaseArmorBonus of the modified item.

### DIFF
--- a/Source/Kesmai.Server/Game/Items/Equipment/Armors/LeatherArmor.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Armors/LeatherArmor.cs
@@ -75,6 +75,7 @@ namespace Kesmai.Server.Items
 			writer.Write((short)2); /* version */
 			
 			writer.Write((byte)_baseArmorBonus);
+			//Do we need to Serialize/Deserialize the Hue?
 		}
 
 		/// <inheritdoc />
--- a/Source/Kesmai.Server/Game/Items/Equipment/Armors/LeatherArmor.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Armors/LeatherArmor.cs
@@ -1,11 +1,19 @@
 using System.Collections.Generic;
 using System.IO;
+using System.Drawing;
 using Kesmai.Server.Network;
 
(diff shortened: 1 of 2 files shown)


Chilly021:
I like this! Great work.

Chilly021 (approved):

trlook:
I'm realizing I should implement the rarity stuff as a dictionary or some other data structure

Source/Kesmai.Server/Game/Items/Rarity.cs
```diff
@@ -0,0 +1,67 @@
...
+	public static string GetRarity(int rarity.Clamp(0,5))
+	{
+		var rareText = "Common";
+		switch (rarity)
+		{
+			case 0:
+				break;
+
+			case 1:
+				rareText = "Uncommon";
```
sproatba, line 15: Two things here, should these be localized?  And should we come up with some more in theme names?  Somthing like worn, fine, exquisite plus a couple more on theme descriptions?

Source/Kesmai.Server/Game/Items/Equipment/Armors/LeatherArmor.cs
```diff
@@ -21,11 +29,28 @@ public partial class LeatherArmor : Armor
 		/// <inheritdoc />
 		public override int BashingProtection => 1;
 		
+		public override int BaseArmorBonus // This is now configured to what we set.
+		{
+			get
```
sproatba, line 34: I could see this going either way.  we may want to be able to give mobs stronger armor without having to equip them with icies, drakes, etc.  right now we only really have 3 armors for mobs (leather, chain, plate).  I could see all 3 getting this treatment simply to give mobs better armor.

sproatba:
looking good

zrikzlok (approved):

PR #1400, opened 2022-03-07
Princess - Removed Jailer

Leolus:
* Removed Jailer
* Added secret door

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -55163,6 +55163,13 @@
         <component type="FloorComponent">
           <ground>3</ground>
         </component>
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

Tawnos-LOK (approved):

starfire5:
Does this include getting a key? The scroll was always questionable to me because we're "crafting" something that already exists. Either way it'll be good to get rid of the jailer, its not needed as Troll King's lair is the jail.

zrikzlok:
looks like this has a conflict needs to be redone due to another commit at some point. may want to pull down latest and then try again

PR #1401, opened 2022-03-07
Balancing changes in Axe

dedmist:
adjusted Mama to not roam on the ledge - this allows for players to strategize as a group
Also reduced Mama spawn timer

Updated prone chances and loot from Giant - there's not really a good reason to go to the Giant, as he's harder than Kosh with a worse payout. He's still harder than Kosh, but at least there's a potion chance now

undead treasure
changed the treasure from the undead area to be specific, not just upper glacier - dropped the lower level items, added low chance of dex - loot more in line with higher level area

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -105295,8 +105295,8 @@
         },
         {
             new CreatureAttack(14,      29, 45,     "The giant stomps on you.",
-                                                        new AttackProneComponent(25), 
-                                                        new AttackStunComponent(10)),               40
+                                                        new AttackProneComponent(15), 
+                                                        new AttackStunComponent(5)),               45
         },
         {
             new CreatureAttack(15,      34, 54,     "The giant slams you across the room.",
@@ -105318,7 +105318,7 @@
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
         new LootPackEntry(true, (from, container) => new ReturningHammer(),        100),
         new LootPackEntry(true, (from, container) => new MoonstoneRing(),        100),
-        new LootPackEntry(true, (from, container) => new PermanentIntelligencePotion(),        50)
+        new LootPackEntry(true, AxePotions,        50)
     ));
        
     return giant;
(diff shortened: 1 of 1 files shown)


Leolus (approved):

Tawnos-LOK (approved):

Chilly021 (approved):

SewerCzar (commented):
This brings up a question I've had for a while: What timer should bosses be on?

Mama is 40 minutes to 1 hour respawn now. Drakes are ~30 minutes.

Should bosses even have a respawn timer if it's meaningless?

dedmist:
Sewer,
At 3x, drakes are 40min-1hr, same as Mama, which used to be at 1-2 hrs; but how many people are lairing in Opal?  Most people lair in a 2x facet, which means we're still looking at 1-1.5 hrs for respawn, which is still longer than OG/NG.

The purpose of a spawn timer was to prevent farming, which really means that the drake spawn timer would be the target for a higher spawn time.  But it shouldn't be too limiting.  It's supposed to be rare enough that it feels like an event, but common enough that it's not a hinderance.

I don't know why you would think this is meaningless, that seems hyperbolic.  Is there a logical or tactical reason that the Mama spawn should be in the 1-2 (1.5-3) hour range?  Is it solving a real problem or is it unnecessary wait time?

dedmist:
You're purposefully ignoring the rest of the post and being hyperbolic again.  Mama isn't being killed or attempted more than a few times a day at peak hours

> I don't see how a 40 minute - 1 hour spawner is rare enough to feel like an event. I was simply asking the logic behind why it should be killable 24-36 times a day instead of 12-24.

Not understanding your comment, how many times a day is Mama being harvested?  Are people sitting up there all day just killing her over and over again, or is she being attempted fairly infrequently at peak times?
The timers were initially set so that when people group hunt they could go across each of the facets without much break and group kill the Trio consistently.  As soon as there's a decent amount of break time the group will disband and not make further attempts.  This was setup to foster further group hunting in the first place, but be far enough a part to force faceting and travel time for additional kills.

So to hear the other side of the story, what's the reason for having very long spawn timers?

SewerCzar:
She actually is currently being killed on spawn timer right now. And no one facets if the timer is that quick - we've seen it over and over on Opal. In my new land we've seen refusal to facet repeatedly and timers were long.

No one is going to facet, no one is going to group hunt her. She's just going to be killed every 40 minutes.

I dont know what the answer is. I was asking what the theory should be originally and I still don't have an argument just a question.

dedmist:
You mean frog kills or using expected game mechanics?  How do we view this data for verification of how these changes are affecting the game?

dedmist:
Once we verify who and why we can look to see what needs to be changed.  Though I definitely haven't seen this since Opal has been a ghost town for the last week.  

I'd be guessing that Mama is too easy and gives too much xp.

PR #1404, opened 2022-03-07
reduce easy scripting areas from leng

dedmist:
add Manticora to drake, adjusted timers
Split drake from minions to ensure respawn timer
added hallway to roam

Bumped down number of concussion wiz and up number of fire wiz to better align with difficulty at level

Reworked waterfall zones with indestructible walls and obstructions, more area coverage and put a mob in the safe zones
doom duck and tiger larger roam

Cleared up deadspot in undead

let power troll roam

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Leng" version="0.90.0.0">
+<segment name="Leng" version="0.92.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


Leolus (approved):

Chilly021 (approved):

SewerCzar (commented):
image

image

image

image

image

image

image

image

image

image

image

image


I'm sure there's more but those are the obvious gaping holes in this

jkachhad:
The best way to prevent scripts might be to reconsider if spawned entities can walk out of their home regions, particularly in non-lairs.

SewerCzar:
> The best way to prevent scripts might be to reconsider if spawned entities can walk out of their home regions, particularly in non-lairs.

yep... even then knights can script in pretty much any zone just in the middle of the spawn.

I'm not sure where this is coming from but if you want it to be no safe plinking zones then you're also removing ALL safe rooms / zones.

I'm just showing spots so once you "fix" them no one comes back crying about how someone else found an exploit spot, like I know they will.

Taking away safe zones seems totally reasonable though. I'm sure MUs would appreciate not having any staging grounds for hunting.

Tawnos-LOK (approved):

PR #1410, opened 2022-03-08
Fix Fire Gobs from burning themselves

Chilly021:


Chilly021:
For whatever reason changes persisted from another branch even though they were checked in. I assume world forge issue.

PR #1413, opened 2022-03-08
Take away DR from a certain shield. Lightning Shield

Chilly021:


Leolus (commented):
Shouldn't we just reduce this to 1 so it's still on par with other shields?

Chilly021:
> Shouldn't we just reduce this to 1 so it's still on par with other shields?

Done, stupid oversight on my part.

SewerCzar (approved):

dedmist (approved):
let the correction proceed

PR #1414, opened 2022-03-08
Hummer Damage Nerf

sproatba:
Attempt to fix some damage issues on hummer based on Discord Discusion

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Player/PlayerEntity.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Player/PlayerEntity.cs
@@ -81,9 +81,10 @@ namespace Kesmai.Server.Game
 			
 			var skillLevel = GetSkillLevel(skill);
 			var skillMultiplier = 1.0 + (skillLevel / skillFactor);
+			var speedMultiplier = weapon.GetSkillMultiplier();
 			
-			var minimumDamage = (int)((baseMinimumDamage + hitAdds) * skillMultiplier);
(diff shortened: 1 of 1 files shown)


Leolus:
At a core level I'm not against the concept here. This will effectively lower the damage of the hummer and increase damage of bows etc. I think we'd want to address this as a weapon overhaul. @dedmist is working on a larger overhaul, should we include this in the plans?

trlook:
I don't think we need to nerf the hummer just yet and can spend some time on how to address the issue. It may be that a larger combat systems pass and itemization pass will address it.

trlook:
Another thing is that flat mitigation double penalizes hummer, so that might also be a balancing knob in the future.

Tawnos-LOK:
> At a core level I'm not against the concept here. This will effectively lower the damage of the hummer and increase damage of bows etc. I think we'd want to address this as a weapon overhaul. @dedmist is working on a larger overhaul, should we include this in the plans?

That was my interpretation as well.  I also like the concept - I'm just curious if additional base range changes would be needed to Hummer and/or bows in combination with this change to get the best overall tuning

trlook:
We might also want to introduce attack speed boosting accessories as a thing.  At any rate, I think there are more things to consider first.

dedmist:
My opinion:
So we have 2 efforts going on, I think making patches on the current game for playability and balancing is a good thing to do, we won't really know exactly what numbers to start with unless we do some tweaking.
I think once we agree up on a land and get it set with the overhaul (which may take some time), then we would need to discuss changes in terms of playstyle and how it affects the game and maybe some area needs an additional mechanic to put it on par.

PR #1418, opened 2022-03-09
Added a new “critical style” attack for mama, changed her to IceDragon class and removed icies from her loot. 

Wirednuke:
I noticed that Mama isn't doing critical damage. Which makes her way easier than TK right now, which is why everyone is running around with icies right now. 

Let me know what you think, wasn't sure if we needed to tone her damage down a little bit to compensate... but I think she does need to crit shot.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106464,7 +106464,7 @@
         <block><![CDATA[]]></block>
         <block><![CDATA[
     //todo add Ice Dragon model / corpse.
-    var dragon = new Dragon()
+    var dragon = new IceDragon()
     {
         MaxHealth = 4200, Health = 4200,
         BaseDodge = 32,
@@ -106529,7 +106529,6 @@
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
-        new LootPackEntry(true, (from, container) => new IceDragonScaleArmor(),        100),
         new LootPackEntry(true, (from, container) => new DragonSkull(),        100),
         new LootPackEntry(true, MamaTreasure,     100.0), 
         new LootPackEntry(true, MamaTreasure2,     100.0),
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106473,7 +106473,7 @@
         
         CanFly = true,
         CanCharge = true,
-        CanStrikeCritically = true,
+        CanStrikeCritically = false,
         BasePenetration = ShieldPenetration.VeryHeavy,
 
         Immunity = CreatureImmunity.Piercing | CreatureImmunity.Slashing | CreatureImmunity.Bashing |
@@ -106493,16 +106493,23 @@
     
     dragon.Attacks = new CreatureAttackCollection
     {
+         {
+            new CreatureAttack(20,      100, 160,      "The Ice Dragon spins around and slams you to the ground with its mace-like tail."),5
+                                                                                                   
+                                                                                                    },
+        
         {
             new CreatureAttack(20,      50, 70,      "The Ice Dragon slashes you with a claw.",
                                                         new AttackProneComponent(5), 
-                                                        new AttackStunComponent(5)),                40
+                                                        new AttackStunComponent(5)),                35
         },
+        
         {
             new CreatureAttack(20,      55, 75,     "The Ice Dragon smashes you against the ice with its tail.",
                                                         new AttackProneComponent(10), 
                                                         new AttackStunComponent(10)),               40
         },
+        
         {
             new CreatureAttack(20,      75, 92,     "The Ice Dragon buffets you with its wings.",
                                                         new AttackProneComponent(15), 
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106494,7 +106494,7 @@
     dragon.Attacks = new CreatureAttackCollection
     {
          {
-            new CreatureAttack(20,      100, 160,      "The Ice Dragon spins around and slams you to the ground with its mace-like tail."),5
+            new CreatureAttack(20,      100, 160,      "The Ice Dragon chomps you with its massive jaws."),5
                                                                                                    
                                                                                                     },
         
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106473,7 +106473,7 @@
         
         CanFly = true,
         CanCharge = true,
-        CanStrikeCritically = false,
+        CanStrikeCritically = true,
         BasePenetration = ShieldPenetration.VeryHeavy,
 
         Immunity = CreatureImmunity.Piercing | CreatureImmunity.Slashing | CreatureImmunity.Bashing |


trlook:
I don't see how critting changes things. If you jump after every hit unless the hit can one shot it doesn't matter.  What should change instead is how fast mama regens out of combat, or resetting her health if she gets leashed out of her lair.

Wirednuke:
> I don't see how critting changes things. If you jump after every hit unless the hit can one shot it doesn't matter. What should change instead is how fast mama regens out of combat, or resetting her health if she gets leashed out of her lair.

Agree with that, just putting her in line with the other major bosses that drop lesser gear.. i.e. TK who crits for 160. 


ALSO, changed her from dragon class to icedragon class so she tans into icies, and removed icies from her loot.

dedmist:
It'd be good to come up with a target way to kill her.  I think she should be a group hunt at 18 and a 50/50 solo chance at 21/17 for a knight.
crit and regen would be a good start, but it'd be nice to come up with a strategy for killing Mama.  Not the same strat as Lucy, but that's a really well done boss.

Wirednuke:
> It'd be good to come up with a target way to kill her. I think she should be a group hunt at 18 and a 50/50 solo chance at 21/17 for a knight. crit and regen would be a good start, but it'd be nice to come up with a strategy for killing Mama. Not the same strat as Lucy, but that's a really well done boss. Not standing in the way of these changes, just some thoughts for the future.

I think that's kind of what i'm getting at... adding the crit chance means there's a random chance for her to destroy you at 18 solo. Having buddies to heal and do additional dps mitigates that risk.

trlook:
I'd rather give here a 5% chance to use an attack that is much much higher damage but doesn't prone and doesn't have a stun component (would use the stun from % of current health instead).  That way you can take a huge hit and have a better chance to react to it.

Wirednuke:
**

> I'd rather give here a 5% chance to use an attack that is much much higher damage but doesn't prone and doesn't have a stun component (would use the stun from % of current health instead). That way you can take a huge hit and have a better chance to react to it.

I don't know how to do that, so if you want to make the changes I'm all for that. This was more or less an attempt to bring to light the fact that I thought she was out of line, and included a proposed change rather than complaining without offering a solution.

dedmist:
Now that I'm thinking about it and want to write it down, I'd be up for having undead ranged adds that thieves/thaums would need to take care of; requires piercing or healing for undead.

trlook:
> **
> 
> > I'd rather give here a 5% chance to use an attack that is much much higher damage but doesn't prone and doesn't have a stun component (would use the stun from % of current health instead). That way you can take a huge hit and have a better chance to react to it.
> 
> I don't know how to do that, so if you want to make the changes I'm all for that. This was more or less an attempt to bring to light the fact that I thought she was out of line, and included a proposed change rather than complaining without offering a solution.

Sure, it's not too hard to do, but I think we should agree on what the target difficulty is on the fight and what the top line damage range should be.

Wirednuke:
> > **
> > > I'd rather give here a 5% chance to use an attack that is much much higher damage but doesn't prone and doesn't have a stun component (would use the stun from % of current health instead). That way you can take a huge hit and have a better chance to react to it.
> > 
> > 
> > I don't know how to do that, so if you want to make the changes I'm all for that. This was more or less an attempt to bring to light the fact that I thought she was out of line, and included a proposed change rather than complaining without offering a solution.
> 
> Sure, it's not too hard to do, but I think we should agree on what the target difficulty is on the fight and what the top line damage range should be.

My opinion: Mama should be a high risk solo at 18, which is mitigated by bringing a healer or more dps.

dedmist:
> **
> 
> > I'd rather give here a 5% chance to use an attack that is much much higher damage but doesn't prone and doesn't have a stun component (would use the stun from % of current health instead). That way you can take a huge hit and have a better chance to react to it.
> 
> I don't know how to do that, so if you want to make the changes I'm all for that. This was more or less an attempt to bring to light the fact that I thought she was out of line, and included a proposed change rather than complaining without offering a solution.

The last number on the right is the percent chance
{new CreatureAttack(20 /*level of attack*/, 90,120 /*range*/, "hit like a Brick" /*no new attack components here*/), 5 /*percentage*/}
Then just reduce one of the others by 5
    dragon.Attacks = new CreatureAttackCollection
    {
        {
            new CreatureAttack(20,      50, 70,      "The Ice Dragon slashes you with a claw.",
                                                        new AttackProneComponent(5), 
                                                        new AttackStunComponent(5)),                40
        },
        {
            new CreatureAttack(20,      55, 75,     "The Ice Dragon smashes you against the ice with its tail.",
                                                        new AttackProneComponent(10), 
                                                        new AttackStunComponent(10)),               40
        },
        {
            new CreatureAttack(20,      75, 92,     "The Ice Dragon buffets you with its wings.",
                                                        new AttackProneComponent(15), 
                                                        new AttackStunComponent(15)),               20
        },
    };

trlook:
I'd say change first attack to be 90-120 damage and remove the stun/prone, change to 5% and move the remaining 35% to the tail smash

Wirednuke:
> I'd say change first attack to be 90-120 damage and remove the stun/prone, change to 5% and move the remaining 35% to the tail smash. She still needs regen though otherwise you can still cheese her no matter what you add until she can one shot.

How does this look now, look at the latest commit.

trlook:
> > I'd say change first attack to be 90-120 damage and remove the stun/prone, change to 5% and move the remaining 35% to the tail smash. She still needs regen though otherwise you can still cheese her no matter what you add until she can one shot.
> 
> How does this look now, look at the latest commit. Do you have a good example of a mob that regens when leashed?

I think the syntax is fine, but maybe have the flavor text be related to biting or chomping.

Wirednuke:
> > > I'd say change first attack to be 90-120 damage and remove the stun/prone, change to 5% and move the remaining 35% to the tail smash. She still needs regen though otherwise you can still cheese her no matter what you add until she can one shot.
> > 
> > 
> > How does this look now, look at the latest commit. Do you have a good example of a mob that regens when leashed?
> 
> I think the syntax is fine, but maybe have the flavor text be related to biting or chomping.

Flavor text updated.

Chilly021:
Honestly I think adding crit back is definitely something that parodies the older stuff. I'll approve.

Chilly021 (dismissed):

Chilly021 (commented):
Actually bunch of stuff has changed. Let me reread this. Please change description.

Wirednuke:
@jkachhad 

I am looking for a way to set Mama's health to full/increase regeneration if she doesn't see a target for x amount of turns. @Leolus and I were talking in discord, and he mentioned it would be possible in segment code, but that seems messy and I feel like this is something we could make use of in other lair situations as well.

dedmist:
Would probably be best to set this in CreatureEntity.cs for all mobs (which we don't have access to), then maybe allow stronger overrides in segments.

Wirednuke:
Should I split out the change to the Icedragon class so it can get put in, and we can talk more about the crit chance stuff if needed?

dedmist (approved):
I haven't done the fight, I think you're right about doing criticals and the ice dragon armor.

However, I'm not entirely sure what the crit equation is?  Are we doing a 2X for crit?  Then there's not an insignificant chance of getting hit for 200-320 during a Mama fight, so instant death.  I am good with this, Mama really shouldn't be in the able to solo category.  Just making sure we're all on the same page for this fight.

sproatba:
she's not being set to crit (2x damage), she's being given one attack that does damage similar to her doing critical damage.

dedmist:
Ok, missed that crit was removed, thanks.

yarrim (approved):
These changes look good to me.

zrikzlok (approved):
I approve of this message

PR #1421, opened 2022-03-09
Make Axe giant drop a random pot every kill

Wirednuke:
Based on a discord voice discussion:

A group of us felt that for the difficulty and skill required to kill the axe giant, his 50% chance of dropping a random pot was a little low.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -105318,7 +105318,7 @@
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
         new LootPackEntry(true, (from, container) => new ReturningHammer(),        100),
         new LootPackEntry(true, (from, container) => new MoonstoneRing(),        100),
-        new LootPackEntry(true, AxePotions,        50)
+        new LootPackEntry(true, AxePotions,        100)
     ));
        
     return giant;


starfire5:
This is a great idea, since most stat pots are random drops throughout the lands at least you're guaranteed to get something good with the giant. Players might get a stat pot they don't often get.

Tawnos-LOK (approved):

dedmist:
It used to be just an int pot, not even a random pot :)

Giant should be close to Kosh difficulty, I'd rather adjust the difficulty or maybe some cash incentive rather than adjust the bottle drop rate.

dedmist:
> This is a great idea, since most stat pots are random drops throughout the lands at least you're guaranteed to get something good with the giant. Players might get a stat pot they don't often get.

At some skill level he's easily farm-able, but also having too many bottles decreases their worth.

Finding items is basically a game based slot machine, it's fun when it finally pays off, but being predictable isn't as engaging.  I get that too low is frustrating, but 50% is still pretty high.

starfire5:
> > This is a great idea, since most stat pots are random drops throughout the lands at least you're guaranteed to get something good with the giant. Players might get a stat pot they don't often get.
> 
> At some skill level he's easily farm-able, but also having too many bottles decreases their worth.
> 
> Finding items is basically a game based slot machine, it's fun when it finally pays off, but being predictable isn't as engaging.  I get that too low is frustrating, but 50% is still pretty high.

I hear you, but stat pots are essential and giant is a more accessible boss that can be solo'd, that players can count on. It may seem predictable but its it's something for us to look forward to. Its more predictable to me to constantly find strength pots when I could really use a willpower or wisdom pot.

Tenscore:
ydnac drops wis.
drakes and serp drops con.
giant drops(dropped) int.

if we're into adding, there could be continuity here via dex, str, and will.   js

Chilly021 (approved):

sproatba:
if the potion is going to be 100% i think i'd add some duds in there like mrp and youth or something.  also maybe nerf the moonstone a bit as well in increase his randomness a bit.

dedmist:
I think we'd be better off removing all of the 100% drop rates for potions except for con.  Then we scale the progression around the expected stat values.  I don't think 15 Giant kills is an appropriate progression path to max out stats, this dulls the lottery effect and makes it more arduous than exciting.
TBH, I think I shouldn't have switched the giant drop in the first place, my thinking was that the Giant is underutilized, but now he's more of an easy win machine at this point - going to 100% drop rate would just exacerbate that.
A 8% drop rate off of Giant level bosses 1 for each stat potion would be a decent way to go, it should be harder than getting potions from cliffs or TT, but easier if you are overleveled for the boss.  We could switch Iatana to a boss fight instead of an escort quest as well - no one ever really does the escort, it was a bad compromise when it went in.

PR #1434, opened 2022-03-12
Fix Shield and Weapon protections

trlook:
Fixed Shield and Weapons to allow for protections to be used.
Added shield and weapon damage mitigation logic to MobileEntity.cs


Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
```diff
@@ -48,12 +48,37 @@ public virtual double GetAttackBonus(MobileEntity attacker, MobileEntity defende
 		/// </remarks>
 		public override int CalculateBlockingBonus(ItemEntity item)
 		{
-			if (item is IWeapon weapon && weapon.Flags.HasFlag(WeaponFlags.Projectile))
-				return ProjectileProtection;
+			var flags = WeaponFlags.Bashing;
```
sproatba, line 51: not sure if this is an issue, but if item isn't a weapon you still return bashing + base armor, should this return 0 instead?

Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.cs
```diff
@@ -22,6 +22,25 @@ public int CalculateDamageModifier(ItemEntity item)
 			
 			if (armor != null)
 				damageModifier += armor.GetArmorBonus(item);
+			
+			/* Calculate armor bonus from held items. */
+			var leftHand = LeftHand;
+			var rightHand = RightHand;
+
+			if (leftHand is Shield shield)
```
sproatba, line 30: should shield work in right hand as well?  i could see someone doing rhammer shield and wanting the / to work for throw over the * or something

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
trlook: initializes as bashing so things that don't use weapons like lairs have a damage type, if the creature weilds a weapon, overwrite that with the weapon flags

Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.cs
trlook: you can just have the raxe in main hand and hsield left and you can use both * and /

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
sproatba: makes sense to some extent, but we are now getting bashed by claws instead of slashed etc.  probably fine in the scheme of things and prevents the need to set damage types on custom attacks

sproatba:
A little worried overall that we are adding too much mitigation, but it may play out ok.  It may smooth things out where a string of bad rolls won't be so penalizing.  Fumble is a bit more devastating now.  I'm not sure if we need to add anything in for status effects here, or up the chain in the code that is calling this.  Being stunned for instance stops weapon blocks, should it stop weapon mitigation?  Also ma gaunts I don't think are handled in here since they are on the paper doll not in the hands.

trlook:
> A little worried overall that we are adding too much mitigation, but it may play out ok. It may smooth things out where a string of bad rolls won't be so penalizing. 
We're not really adding any mitigation, this should only affect projectiles and only with shields and OL staff.  It opens up the opportunity to put some protections on future items.

>Fumble is a bit more devastating now. I'm not sure if we need to add anything in for status effects here, or up the chain in the code that is calling this. Being stunned for instance stops weapon blocks, should it stop weapon mitigation? 
 I don't think the mitigation needs to be stunned away, it could be a magical property of the item. 

>Also ma gaunts I don't think are handled in here since they are on the paper doll not in the hands.
No gauntlets currently offer protection, nor any thing like rings. If we wanted to add that in the future, it's easy to extend. The reason this needs to go in now for weapons is because a protection was added to the OL staff.

trlook:
> > A little worried overall that we are adding too much mitigation, but it may play out ok. It may smooth things out where a string of bad rolls won't be so penalizing.

 We're not really adding any mitigation, this should only affect projectiles and only with shields and OL staff.  It opens up the opportunity to put some protections on future items.

> > Fumble is a bit more devastating now. I'm not sure if we need to add anything in for status effects here, or up the chain in the code that is calling this. Being stunned for instance stops weapon blocks, should it stop weapon mitigation?

 I don't think the mitigation needs to be stunned away, it could be a magical property of the item.

> > Also ma gaunts I don't think are handled in here since they are on the paper doll not in the hands.

 No gauntlets currently offer protection, nor any thing like rings. If we wanted to add that in the future, it's easy to extend. The reason this needs to go in now for weapons is because a protection was added to the OL staff.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
```diff
@@ -53,32 +77,47 @@ public override int CalculateBlockingBonus(ItemEntity item)
...
+			{
+				weaponBonus += ProjectileProtection;
+			}
+			else
+			{
+				weaponBonus += GetMeleeBonus(item, flags)
 			}
 
-			return blockingBonus + BaseArmorBonus;
+			return weaponBonus;
```
sproatba, line 118: this seems kind of inconsistent to me, If we are going to add the mitigation in there might as well be consistent,  Adding it in though might require some balance changes overall though (and adding in gauntlets).

trlook: That there was because I didn't think something like the BBS should provide 3 damage mitigation. Also why I broke up the DR and Mitigation portions...to make things more consistent

jkachhad:
To understand what we are doing.

- `CalculateDamageModifier` now calculates mitigation provided by armor, shield, and left/right weapons?
- `GetMeleeBonus` on Weapons now provides a mitigation value. `GetShieldBonus` on shields provide a mitigation value.
- `CalculateArmorBonus` now only uses `BaseArmorBonus` and is only used to determine blocks.

trlook:
> To understand what we are doing.
> 
>     * `CalculateDamageModifier` now calculates mitigation provided by armor, shield, and left/right weapons?

Correct, or more accurately it sums the calculations from `GetMeleeBonus`, `GetShieldBonus`, and `GetWeaponBonus`. For weapons, the right hand only applies if the left hand is empty, or if the left hand has a shield and the right hand is not a two-handed weapon. Currently, the only items that make use of this are shields and OL staff and only through projectile protection. I put in the code for the other protections in case new items providing these protections are added.

>     * `GetMeleeBonus` on Weapons now provides a mitigation value. `GetShieldBonus` on shields provide a mitigation value.?

`GetMeleeBonus` is a helper method in both Weapons.cs and Shield.cs to do the attacker weapons flag check since it was needed in both `CalculateBlockingBonus` and `GetShieldBonus`/`GetWeaponBonus`. As stated above, the mitigation values are pulled from `GetShieldBonus`/`GetWeaponBonus` in `CalculateDamageBonus`

>     * `CalculateArmorBonus` now only uses `BaseArmorBonus` and is only used to determine blocks?

Not sure I touched a `CalculateArmorBonus` function. `CalculateBlockingBonus` has differing behavior for Shields and Weapons. 

For weapons, only `ProjectileProtection` grants block against projectiles. Having the `BaseArmorBonus` contribute to projectile bonus would be too strong and would have buffed something like the BBS. This was the previous behavior and left unchanged. Added was to allow Slash/Bash/Pierce protection (of which no weapon currently has) to contribute to melee blocking (for possible future use).

For shields, allowed `BaseArmorBonus` to contribute to projectile blocking, would likely need to revert Brick's buff of adding +1 `ProjectileProtection` to each shield. We could also just mimic the weapon behavior here instead.

To summarize:

- Allow `Protection` properties on weapons/shields to mitigate damage provided proper wielding conditions are met.
- Allow `Protection` properties to contribute to Blocking on shields and weapons (I assume the Blocking has wielding conditions checked in a file that is not exposed). Allow BaseArmorBonus to contribute to blocking projectiles for shields but not weapons (but open to disallow for both).
- Currently only affects Shields and OL Staff, but future items making use of these properties is enabled.
- Armor behavior untouched.

jkachhad:
I've refactored some naming conventions to make it easier to understand in the future. "Modifier" was confusing, so I renamed it to Mitigation. 

1. 
I think breaking up `BaseArmorBonus` to contribute specifically to blocking (AR vs DR) is a good move forward for item balancing. However, I don't think we should use `BaseArmorBonus` as part of `Protections` in any part of the calculation.   When balancing items in the future, we/someone may not realize that `ProjectileProtection` dips into `BaseArmorBonus` for the final mitigation value.

We should instead, for ranged and overlap staff, modify the raw value and add `BaseArmorBonus`.  It's than easier to think, "if the attack is mitigated, here is the mitigation"  rather than it's mitigated by Protection + BaseArmorBonus for Shields, but it's different for other weapons.

2.
If you agree with the above, we should rename "Protection" to "Mitigation" across the project.

trlook:
- Agree that `Protection` should be renamed to `Mitigation`. (Might consider for things like Fire/Ice where it reduces damage to distinguish from S/D Protection)
- `BaseArmorBonus` should only provide block and needs to be altered in Armor.cs as well.
- Rename `BaseArmorBonus` to `BaseBlockBonus` for clarity.
- Should Mitigation also give DR or should be break the current Protections up further?

I really only had BaseArmorBonus adding to mitigation on shields because that's how it was implemented for armors. I realized as I was coding that it would be OP or weapons, hence the exception. I agree it's better if BaseArmorBonus never mitigates, and we should be consistent across all items with that.

Previously, all armors only used Protections so BaseArmorBonus was never actually used there.  However, with quality on armor defaulting to adding BaseArmorBonus, the question "Do we want armor quality to only block more or block AND mitigate more or only mitigate more?" arises.

dedmist (approved):
Having these being more granular is a benefit.  I'm good with the shield mitigation, but I think that the OL staff shouldn't really offer projectile mitigation nor do I think it should work with a shield.  This kindof makes the OL staff a "best" weapon and removes choice.

Chilly021 (dismissed):

Tawnos-LOK:
I think I understand the intent behind this PR without fully understanding the impact(s).  My gut feeling is that it will lead to a whole new round of itemization tweaks, which may not be a bad thing if the core is healthier.

Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.cs
```diff
@@ -6,24 +6,51 @@ namespace Kesmai.Server.Game
 	public abstract partial class MobileEntity : WorldEntity
 	{
 		/// <summary>
-		/// Calculates the damage modifier against the specified item.
+		/// Calculates the damage mitigation against the specified item.
 		/// </summary>
-		public int CalculateDamageModifier(ItemEntity item)
+		public int CalculateDamageMitigation(ItemEntity item)
```
sproatba, line 11: I think the code to actually call the gauntlet functions is missing here still.

Source/Kesmai.Server/Game/Items/Equipment/Gauntlets/Gauntlets.cs
```diff
@@ -137,6 +139,22 @@ public override int CalculateBlockingBonus(ItemEntity item)
 			return BaseArmorBonus;
 		}
 		
+		/// <summary>
+		/// Gauntlets at this time do not provide mitigation against weapons.
```
sproatba, line 143: I think we should go ahead and add the code, that way if we need to tweak gaunts later we don't also have to revisit this in order to do it. Setup the knob to turn even if we aren't currently turning it.

Source/Kesmai.Server/Game/Items/Equipment/Shields/Shield.cs
```diff
@@ -54,18 +55,74 @@ protected Shield(int shieldID) : base(shieldID)
...
+			if ((flags & WeaponFlags.Projectile) != 0)
+			{
+				shieldBonus += ProjectileProtection;
+			}
+			else
+			{
+				shieldBonus += GetMeleeBonus(item, flags);
+			}
+
+			return shieldBonus + BaseArmorBonus;
```
sproatba, line 123: Name refactoring still seems a bit off but I believe this should not add baseArmorBonus

jkachhad:
I've added some additional refactoring. The first push renames all Protection to Mitigation.
The second changes combat calculations, separating DR and Mitigation calculation.

DR does not depend on any Mitigation values. Mitigation does not depend on BaseArmorBonus.

jkachhad:
```
/// <inheritdoc/>
		public int CalculateMitigationBonus(ItemEntity item)
		{
			var itemProtections = new List<int>();

			/* Determine the type of weapon attacking us. For hands, we use bashing damage. */
			var flags = WeaponFlags.Bashing;

			if (item is IWeapon weapon)
				flags = weapon.Flags;
			
			/* Determine mitigation. */
			var protectionBonus = 0;
			
			if ((flags & WeaponFlags.Projectile) != 0)
				itemProtections.Add(ProjectileMitigation);

			if ((flags & WeaponFlags.Piercing) != 0)
				itemProtections.Add(PiercingMitigation);

			if ((flags & WeaponFlags.Slashing) != 0)
				itemProtections.Add(SlashingMitigation);

			if ((flags & WeaponFlags.Bashing) != 0)
				itemProtections.Add(BashingMitigation);

			if (itemProtections.Any())
				protectionBonus += itemProtections.Min();

			return protectionBonus;
		}
```

Is now pretty common among anything that implement `IArmor`  I could not see any issues as it is offhand.  Mitigation should be equal across Armor, Weapon, Shield, and Gauntlets?  Are there any cases where any of those many not work against specific weapon flags?

jkachhad:
I could now see the need to add additional modifier for `CalculateBlockingBonus` to Shields against projectiles.

- Do we use the mitigation value?  They would essentially get `(BaseArmorBonus + ProjectileMitigation)` - which makes it a bit odd.  It would double dip into mitigation and break the separation.
- Do we add a general rule that shields provide +2 `BaseArmorBonus` against projectiles? This makes it difficult to configure as larger shields should offer better bonuses.
- Do we create an additional property for `Shield` only that is `ProjectileArmorBonus` ?

jkachhad:
After playing around some, before this can be merged in, we need to look at `BaseArmorBonus` across all `IArmor`.  Typically it should I think be minimum of the Mitigations.

LeatherArmor => 1
ChainmailArmor => 2 or 1 (I would argue for 2, can't be used by MUs, benefits fighters with low ranged defense)
TrollVest => 2
WyvernScales => 1
Platemail => 2 (I would argument for 3, same as above)
SalamanderScales => 2
DragonScaleArmor => 3
DrakeScaleArmor => 4
IceDragonScaleArmor => 5

jkachhad:
Here would be my new suggestion of BaseArmorBonus, Mitigation, and Quality interacting.  
https://docs.google.com/spreadsheets/d/1fzJdWJj7X-pf8ER6xZ2o5W7wbrFJ7TmFpXTtNw8heL0/edit?usp=sharing

As a general rule, maybe we should make it so quality only ever changes 1 property of an item to simplify game mechanics?

Armor: BaseArmorBonus
Weapons: BaseAttackBonus
Shields: ProjectileProjection
Gloves: Maximum damage

trlook:
> I could now see the need to add additional modifier for `CalculateBlockingBonus` to Shields against projectiles.
> 
>     * Do we use the mitigation value?  They would essentially get `(BaseArmorBonus + ProjectileMitigation)` - which makes it a bit odd.  It would double dip into mitigation and break the separation.
> 
>     * Do we add a general rule that shields provide +2 `BaseArmorBonus` against projectiles? This makes it difficult to configure as larger shields should offer better bonuses.
> 
>     * Do we create an additional property for `Shield` only that is `ProjectileArmorBonus` ?

Sorry, been out of town so just catching up on this. That was sort of the conundrum I got into when first working on this fix. I can see wanting shields and armors to "Block" projectiles but weapons not being able to. And what about the OL staff, is it just mitigating projectiles or helping block them?

I also want to propose a weapon flag of "Versatile" or the like where it can be wielded one or two-handed to separate out the true two-handers like the SGA, Halberd, Staff, etc. from things that might be used with a shield like BBS and Spears.

trlook:
> Here would be my new suggestion of BaseArmorBonus, Mitigation, and Quality interacting. https://docs.google.com/spreadsheets/d/1fzJdWJj7X-pf8ER6xZ2o5W7wbrFJ7TmFpXTtNw8heL0/edit?usp=sharing
> 
> As a general rule, maybe we should make it so quality only ever changes 1 property of an item to simplify game mechanics?
> 
> Armor: BaseArmorBonus Weapons: BaseAttackBonus Shields: ProjectileProjection Gloves: Maximum damage

I like this. Maybe in the future we can have additional properties that can be attached to an item that scale off quality, or skills/spells that check weapon quality when they are used/cast.

trlook:
> Long PR here. Looking at a month long open now. Should we close it or should we re-open under a different context?

I don't think this is ready, but it's being worked on... I'm not sure what the best work flow is tbh. I opened the PR so people could see and discuss, but there are a lot of entangled issues.  If we can move all the work to a specific branch and have a discussion thread about it there that would work, otherwise we could leave this open and keep developing the branch here.

sproatba:
I've been through this, and I believe what is here can go and work, but some pieces of the puzzle are missing so lux would need to review how it affects some of the hidden combat calcs.   I'm not sure I really like that we took all the control we had over how things blocked and reduced it to a single number for all types of damage in favor of being able to reduce on hit damage from all types.   I'm not sure that is really a winning scenario.  Previously we had a nice blocking progression on shields for archers, this change makes the shields all block the same, but each mitigates slightly more damage.

The fan being the outlier that it is, still needs looked at in the context of this as well.  I don't think it's working as intended and just gets far worse with this implementation.

jkachhad:
So, then should armors double-dip into this number?  It would be used for both a DR calculation and Damage mitigation.

It makes the requirement to hunt a new area high, which may be not a bad thing.  You have to have gear that ensures you don’t get hit,  it also mitigates damage.  Makes it difficult for MAs though at high difficulty levels as they have no additional block.

dedmist:
Mitigation for armor seems like it's going to make balancing more difficult. MAs will fall behind and the effect for knights will be exacerbated. Say the block rate is 75% and the damage averages 15, with ice dragon scales,  this turns the effective block rate up to 82%, (27% damage reduction) without even taking into account the bonus armor. That's an effective increase in HPs, which is good for MU, but unbalanced for knights.

I would put mitigation on MU only items/spells (maybe +1 hat, +1 ring, +1 robe or an imbue).
This would close the gap a bit without killing the balance.  This shouldn't be a tradeoff for MUs.
Some armor mod and arrow mitigation is still good for shields.  MAs could have arrow mitigation through an item - a bracer or having to turn neutral for a fan.  This should be a tradeoff for melee.

Then for dynamic mitigation, for flavor, I would drop the solid block rate X%, add a glancing hit X+Y% of the time with some mitigation.

Balancing for anything below dragonscale only affects the first hour of gameplay, and I wouldn't expect anyone to do armor switching to optimize mitigation.   Slotting between dragon and drake or after ice scales would be more useful.

Chilly021:
To my understanding @jkachhad is looking at this holistically. We can re-open this if we need to use this for staging other changes.

Chilly021 (commented):
Re-opening as requested, but a lot of merge conflicts to resolve. Might be easier to open a new PR?

n099y:
@jkachhad is this still something you would endorse ?
I do like the idea of Mitigation being added to armors. If your still on board I can devote some time to bring this back to life.

Chilly021:
We can come back to this if needed! Definitely awesome conversation piece. Though the PR looks like a LOT of conflict resolve right now. Might be best to re-do this one in a separate PR. 

@trlook

PR #1445, opened 2022-03-14
Changes to Misericorde Dagger and how it drops

SewerCzar:
Proposal to change Misericorde dagger to a less-powerful version of the BBS (Medium armor pen compared to Very Heavy, 1-10 damage instead of 2-12).

It will drop from yeti and bind to the person who gets the killing blow so no recall for twinking people.

It will also not be a returning weapon.

I believe there will be zero impact to balance but will make dagger players happy.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106596,8 +106596,7 @@
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
         new LootPackEntry(true, AxeGems,       100,     gemsPriceMutatorVeryHigh), 
        new LootPackEntry(true, YetiTreasure,     100.0),       
-       new LootPackEntry(true, (from, container) => new ManaPotion(),        100),
-       new LootPackEntry(true, (from, container) => new MisericordeDagger(),        100)
+       new LootPackEntry(true, (from, container) => new ManaPotion(),        100)
    ));
     
    return yeti;
@@ -106609,6 +106608,12 @@
         <block><![CDATA[
     if (killer is PlayerEntity player)
     {
+    
+        var backpack = source.Backpack;
+    
+        if (backpack != null)
+            backpack.AddItem(new MisericordeDagger(player));
+            
         var quest = Quests.GetQuest<KnightsDragonSlayerQuest>();
     
         if (quest != null && quest.InProgress(player))
--- a/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Daggers/MisericordeDagger.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Daggers/MisericordeDagger.cs
@@ -11,7 +11,19 @@ namespace Kesmai.Server.Items
 		public override uint BasePrice => 30;
 
 		/// <inheritdoc />
-		public override int BaseAttackBonus => 3;
+		public override ShieldPenetration Penetration => ShieldPenetration.VeryHeavy;
+
+		/// <inheritdoc />
+		public override int BaseArmorBonus => 3;
+
+		/// <inheritdoc />
+		public override int BaseAttackBonus => 4;
+
+		/// <inheritdoc />
+		public override int MinimumDamage => 1;
+
+		/// <inheritdoc />
+		public override int MaximumDamage => 10;
 
 		/// <inheritdoc />
 		public override WeaponFlags Flags => base.Flags | WeaponFlags.BlueGlowing | WeaponFlags.Neutral;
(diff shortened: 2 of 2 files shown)


Chilly021 (approved):

zrikzlok (dismissed):

Tawnos-LOK:
Zero impact to balance?  It would make the dagger with the strongest defense also have the most damage range and attack adds with Very Heavy armor pen.  The Backstab damage with two would be entirely off the charts and be effectively indefensible but maybe that's the real point here.

starfire5:
Wouldn't that be overpowered? A dagger is primarily a thieves weapon and a thief could stack their damage if they have the poison spell.

Chilly021 (changes requested):
I could see this a bit more on main gauche. I think brick put something similar in on the other lands.

https://github.com/jkachhad/Stormhalter/blob/master/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Daggers/MainGauche.cs

trlook:
Correct me if I'm wrong, but ArmorPen basically has no benefit to players (besides adding damage to hand attacks with gauntlets) since no creatures have shield spell? I'm actually pro adding a dagger blocking weapon, but maybe not so offensively powerful. Brick and I discussed something like just adding 2 BaseArmorBonus to Yeti Dagger would make him happy.

SewerCzar:
> Zero impact to balance? It would make the dagger with the strongest defense also have the most damage range and attack adds with Very Heavy armor pen. The Backstab damage with two would be entirely off the charts and be effectively indefensible but maybe that's the real point here.

It doesn't have VeryHeavy armor pen, and 1-10 damage range instead of 2-12 or 4-14. It also has less attack adds than rdagger.

It's still less powerful than a BBS, and drops from an endgame boss you'd have to get the killing blow on.

SewerCzar:
> Correct me if I'm wrong, but ArmorPen basically has no benefit to players (besides adding damage to hand attacks with gauntlets) since no creatures have shield spell? I'm actually pro adding a dagger blocking weapon, but maybe not so offensively powerful. Brick and I discussed something like just adding 2 BaseArmorBonus to Yeti Dagger would make him happy. Thieves recently buffed through armor mitigation, so this might be a bit overboard.

Not sure on what the armor mitigation buff is. Is that the change to player armor?

In reality this dagger does less damage than an raxe, has less shield pen, isn't returnable, and off an end game boss that you'd have to get killing blow on.

ArmorPen is more regards to people getting ganked, hence medium armor pen.

Tawnos-LOK:
It has very heavy shield pen, it's in the commit.  Not sure why you keep saying it's medium.

SewerCzar:
> 

I agree with a lot here. Hm. I guess the main problem is thieves right now basically don't have bow as an option (super low DPS, extremely unsafe hunting environments). They have MA skill, which makes them a squishy MA with less damage, and then they have rdagger with fan offhand. 

Problem with making this +4 with low block is now it's a main hand weapon with no returning that will also just be used with a fan until rdagger is gained... And would be as hard or probably harder to get than rdagger.

I'm thinking that it's being compared against other daggers too heavily, not against other items in general.

SewerCzar:
> This is pretty close to what Brick was asking for, but I think I need to understand what the target playstyle would be. As an attacking dagger, I'm all in for Medium Pen and +4 adds. I think people should make a tradeoff though, for blocking.
> 
> game play opinion: So, in a dual wield thief, made for toe to toe, I'd like to have a blocking dagger, like the main guache and either a misericorde or a rapier for attack. The rapier should be the superior weapon (esp since it can't backstab), the misericorde should cover the ground with a poison, but fall behind in the long run.
> 
> What I think wouldn't be good is a misericorde and a rdagger which is where I would disagree with the armor bonus.
> 
> Onboard: +4, adjust 1-4 range if necessary, medium-heavy pen; this is a forced toe to toe attack or a backstab multiplier with a second weapon, which has more limited usage. I'm fine with the high burst damage aspect of this, wouldn't even be offended if there's a TK/drake strategy here that's faster but involves more danger.
> 
> Not onboard: higher blocking, very heavy pen.
> 

Reading this again has me wondering a lot of shit. Probably best left to post-item rework from the ground up.

I'm viewing this as "it's a less-good BBS but thieves need a way to hunt in grounds without having zero defense or a fan." Or a "thieves have backstab they need daggers they can dual wield and hunt with." 

I tuned it down to 2 defense / 4 attack with medium. It'd be a good dual wield dagger, good for backstab, and requires a killing blow on an end game boss that is not simple in the slightest for thieves. I think my tuning is very reasonable but it doesn't actually address any of your concerns.

dedmist (approved):
Ya, let's try it out, we can re-investigate when we do something with the main gauche and the rapier.

trlook:
Brick was saying that the r dagger is a strong main hander and just needs a blocking off-hand dagger.  +4 attack seems fine enough offensively and +2 base armor seems like a good starting point on par with mace offhand defensively. I’m not sure about the damage range but I do support making dual dagger thieves more viable. My proposed weapon “protection” changes also introduce some defense options beyond just basearmorbonus.

Wirednuke:
I'm not against this, it lends itself well towards having multiple flavors of dps thief. 

Misery in offhand, Rdag in main hand - Slightly improved blocking over the other daggers, hard to argue with +5 attack rating on the rdag even for melee. This would be a higher burst damage way of rolling with backstab. 

Main Gauche Offhand, Rapier in main hand - Much better blocking, more of a toe to toe fast melee build.. pretty similar to the way hand is played right now with fan/dagger. 

I like having more options, as long as there are tradeoffs between strengths and weaknesses between the options.

Leolus (approved):

PR #1454, opened 2022-03-17
Experience tweaks on Sewers, Speedy and -300

Wirednuke:
After reviewing the xp settings in these areas I understand why the spawn modifiers were set in the sewers. Both -100 and Sewers share the same mob table, so the xp modifier in sewers was placed to make them give a little bit more experience without having to make another set of mobs. I decided to continue using the experience modifier in sewers, but made the following changes:

- Adjusted the XP modifier in sewers from 2.2 to 1.8 - This brings the xp in line with areas of similar difficulty (TT, Ianta Leng, and Griff Cliffs Leng)
- Raised the XP of the shared -300/Speedy mobs by a modifier of 1.2. This was changed directly in the entities tab. This should put both Speedy and -300 in line with the xp/hour gained in sewers currently.

Darkice worked on these changes with me, and agrees with them as well.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Underkingdom" version="0.92.0.0">
+<segment name="Underkingdom" version="0.93.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

sproatba:
Competing pull request here and one sewer did removing the xp modifiers.  Need to decide if xp modifiers are the way to go or if they should all be set to 1 and levels adjusted through entities (or the mob stuff asheron is trying out).

dedmist:
modifiers were a gap filler, if we design the creatures for the zone then the modifiers are unnecessary.  The zones have been balanced based on a progression instead of the modifiers, so these should be removed in favor of #1455

Leolus:
Closing this due to the conflict mentioned above, reopen a new request if we still believe changes are required after testing.

PR #1455, opened 2022-03-17
Fixed an issue where experience modifier artifacts remained

SewerCzar:
Sewer modifier was in when we were originally using the -100 tunnels monster values. With new system experience modifiers are unneeded (but should we go with Kora's ideas of weighting zones for gold/xp/loot balance we can edit as needed).

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -174,7 +174,6 @@
       <id>1</id>
       <name>Surface</name>
       <height>0</height>
-      <experienceMultiplier>1.1</experienceMultiplier>
       <level>5</level>
       <tile x="0" y="0">
         <component type="ObstructionComponent">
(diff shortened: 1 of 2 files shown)


dedmist (approved):
Agreed, we have finer grained controls and shouldn't need these unless we're doing something special.

sproatba:
without some addition changes, this just reduced the difficulty of sewers and reduced experience as well.  Is that the goal?

dedmist:
These areas were already changed with increased experience and normalized hp, with the experience modifiers they're too high now.  There's another pull request that confirms the xp from the lightning caves is too high.

zrikzlok (approved):

Tawnos-LOK:
As @sproatba mentioned in #1454, there are two approaches being advocated for in competing PRs - with modifiers and without

Chilly021 (approved):

PR #1460, opened 2022-03-18
Kesmai overhaul: part 1, consistency

dedmist:
What's being done here?  I'm essentially creating the framework for creature tables, this will allow us to build a progression for each creature, so that we can balance area progressions across all lands.  BUT***  I'm not making major changes just yet, first thing is to get the framework in and make sure people like the approach.

***WARNING***
I created a test project to compile all of this, but I may have made a mistake somewhere and caused a compile issue.  Let's take this into account when merging and consider that it might break things, so may need to be backed out and fixed.

This is a PoC: there is new infrastructure which is currently in the Kesmai segment, though I'll probably break this out into a MonsterManual class so that it can apply across all lands.

So there's a ton of file changes, but they're mostly infrastructure, no balancing changes yet.

What I've created is a creature table generator and 2 test cases on the surface as far out as I can get, those should be examples of how we can create creatures going forward, it reduces 90% of the code and allows us to ensure consistency.

The purpose of this first change is a PoC to make sure it all works before I get too deep.


sproatba:
I think something like this could go a long way although i'm not sure if this is the best way to accomplish it.  I'd almost rather see more of this baked into the individual classes and you pass in a level.  that way you don't need to add a bunch of attacks to wolf in every land a wolf exists in.  If you wanted a boss wolf you could add or override.  Lots of stuff should just be common across lands and the differences should be mainly what loot they have and how strong they are.  Definitely worth exploring some way of cutting down on the repetitiveness of stuff redefined in each land.

dedmist:
Agreed, the intention is to handle all the trash mobs this way, so the attacks would be generated in the Behavior delegate for custom attacks.  Right now only the default attack is being generated as a test.
There are enough commonalities between different creatures that a progression builder makes more sense than baking it directly into the class for most things.  Though we can still do things like push NV onto bears in the base class instead of generating that in the behavior.
Right now it's in the kesmai segment to see if people like it, then I'll move it to a MonsterManual class at the same level as the CreatureEntity class and it will be usable across all the segments.

Chilly021:
If you are just testing it for future proof of concepts, lll just approve. We just need to be explicit on the changed to users.

Chilly021 (approved):

SewerCzar (commented):
This one's above my skill set so I'll sit by. I'd like Lux to look at this preferably but if it goes in it's all good.

dedmist:
@SewerCzar 
Essentially it's a monster manual that roughly follows your spreadsheets for creature generation, so that we can standardize all of our creature generation and target difficulties.
There are only really 2 things to understand:
1: This is a creature definition, there are a number of default parameters and equations that can be set for the level progression, we can even make them non-linear if we want, we just define new equations for the specific class.
`[typeof(Bear).Name] = new MobMutator {healthBase=13, dodgeBase=12, damageBase=10, damagePerLv=1.5f, attackBase=3, goldBase=30, goldPerLv=6, expBase=850}`
2: Here's an example usage.  The original underlying class can be manipulated the current way, but it can also be completely default based on the monster manual.
```
    var bear = generateMob<Bear, AnimalBehavior>(3, Area.kesmaiSurface);
    bear.Movement=2;
    bear.AddStatus(new NightVisionStatus(bear));
	return bear;
```

zrikzlok (approved):
I’m curious to give this a try

PR #1464, opened 2022-03-20
Fixing an exploitative location

SewerCzar:
Location is being used as an unintended safe zone. Necessary fix is in.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Axe Glacier" version="0.93.0.0">
+<segment name="Axe Glacier" version="0.92.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
@@ -110416,6 +110416,7 @@
       <entry entity="level16Worg" size="3" minimum="2" maximum="2" />
       <bounds region="2">
         <inclusion left="-2" top="1" right="27" bottom="30" />
+        <inclusion left="17" top="-3" right="27" bottom="0" />
         <exclusion left="18" top="23" right="27" bottom="30" />
       </bounds>
     </spawn>


dedmist (approved):
Agreed, this was an uncovered area in Axe UD.

Chilly021:
I still am scratching my head in regards to safe areas and the word exploitative. I don't want to get super reactive about these things going forward for the sake of people needing to get to conf or not. Recall pulling and such is a bit different. I feel like with the latest changes newer people or more casual people are getting the short stick. I am not saying this change does not need to happen, but this does feel more targeted. Let's just be careful to not keep calling everything a dang exploit if it was the original design. More issues are coming from thief areas that literally you could walk away and come back and still be alive, just because of thief hide and the areas are not extended and allow plinking. In this scenario people were pulling back to safety. Getting to that area I assume is pretty hard (again assuming). I hear the need to remove any need to exploit, but will there be safe areas going forward.

dedmist:
If we're going to build in safety rooms/zones, then they should be purposeful and considered.  The areas that are being covered up now, are areas where mobs don't walk through, but also have lots of visibility and no real danger under a large variety of circumstances.

What does a good safe room look like?  Limited visibility, indestructible walls, a secret door.  Or, destructible, but where mobs will enter the area.  These can be used to drag mobs into for control.  There's enough danger that unattended scripting is more difficult to do.  

Areas separated by empty passages are good safe rooms.

Areas that allow for someone to sit all day casting WW/PC while collecting experience without ever having to deal with danger in the highest xp areas of the game would be epitome of bad game design.

The point is to reward actually playing the game, so if your playstyle is to sit and plink, you can still do that, just not without dealing with a dangerous mob from time to time and it should be less efficient than fighting in the dungeon.  If you want to sit and cast ww/pc all the time, you can still do that, but again, not without some danger from time to time and the efficiency should be lower.  But we shouldn't cater to these styles, they're less engaging.

Can we have some regrouping areas that are relatively safe? Sure, like the ledge south of Mama, the entrance to OL, the ledge next to Vlad's tower, the room before shid's Ninjas, these are all well done areas and facilitate grouping.  A safe room, with limited visibility close to the AxeUD boss would be valuable, maybe even a connecting tunnel to AxeUD.

Feel free to point out the specifics of newer and casual players having survivability problems and we can fix them.  I think the -300 gargs probably need to be adjusted down just a bit, their poison hits a bit hard.

sproatba:
If you can walk away in complete safety, or gain experience and skill with zero risk it's exploitable.  It's ok to have a safe room where you open the door and drag stuff in, but occasionally something should spawn there unless it's a large no recall zone or something where it makes sense to give a safe haven.   

There is also the issue of transition between levels to address.  There is a huge difficulty jump into this area, so something to compensate for that should be available, right now it's the ability to make it to the safe area that makes this area huntable as a next step after other axe areas, removing that without some amount of compensation just makes this yet another noob trap.  I can hunt the area before this just fine lets try the next one, oops i'm dead the second you go down the steps.

SewerCzar:
> I still am scratching my head in regards to safe areas and the word exploitative. I don't want to get super reactive about these things going forward for the sake of people needing to get to conf or not. Recall pulling and such is a bit different. I feel like with the latest changes newer people or more casual people are getting the short stick. I am not saying this change does not need to happen, but this does feel more targeted. Let's just be careful to not keep calling everything a dang exploit if it was the original design. More issues are coming from thief areas that literally you could walk away and come back and still be alive, just because of thief hide and the areas are not extended and allow plinking. In this scenario people were pulling back to safety. Getting to that area I assume is pretty hard (again assuming). I hear the need to remove any need to exploit, but will there be safe areas going forward.

This is a location you can WW/PC from with zero threat. And pull mobs into with no threat of pulling more than the intended number of packs. There's no surprise spawn it's a spot that can be abused by any class.

Chilly021 (approved):

zrikzlok (approved):

Tawnos-LOK (approved):

PR #1465, opened 2022-03-20
Jarfel, Carfel's evil swashbuckling cousin

Wirednuke:
This is just a draft pull. Just an idea for implementing the Main Gauche for testing. I made a small sub-lair behind Shidosha that is accessible only by thieves. Jarfel is visiting Mr. Shidosha for ninja training.  It's pretty much a Carfel cut and paste job.  It doesn't have to be permanent, but it's simple and works for now. 

My alternate idea was to create Jarfel a lair in Leng undead, but that was a lot more intrusive. 

If you guys like it, I can set it ready for review and we can push it in.

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Leng" version="0.92.0.0">
+<segment name="Leng" version="0.93.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


dedmist (approved):
I think what we need to ultimately do is come up with a new lair in Oak for this.
Having said that, this is a fine solution for now, we need to be able to play test a blocking dagger.

Wirednuke:
> I think what we need to ultimately do is come up with a new lair in Oak for this.
> Having said that, this is a fine solution for now, we need to be able to play test a blocking dagger.

A few of us talked about this today, and I think I’m going to close this pull and add the blocking dagger to the liches in oak UD

PR #1466, opened 2022-03-20
Add Max Range to see if this actually is usable to KHally

Chilly021:
Add Max Range to see if this actually is usable to KHally

### DIFF
--- a/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Halberds/CopperHalberd.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Halberds/CopperHalberd.cs
@@ -24,6 +24,8 @@ namespace Kesmai.Server.Items
 
 		/// <inheritdoc />
 		public override int BaseAttackBonus => 5;
+
+		public override int MaxRange => 2;
 		
 		/// <inheritdoc />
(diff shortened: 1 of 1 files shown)


trlook:
I suspect there would need to be client tweaks to make this work

Chilly021:
Quite possibly. I did see the property exposed on the props. So it has me wondering.

dedmist (changes requested):
I only see MaxRange being implemented in the Gauntlets base class that doesn't inherit directly from Weapon, so I'm not entirely sure that this is valid.  Also note that Lux is using an annotation on Gauntlets.  I'm not sure what he's using the annotation wrapper for, but I'd check on the validity and function first.

		[WorldForge]
		[CommandProperty(AccessLevel.GameMaster)]
		public int MaxRange => 0;

Chilly021:
> I only see MaxRange being implemented in the Gauntlets base class that doesn't inherit directly from Weapon, so I'm not entirely sure that this is valid. Also note that Lux is using an annotation on Gauntlets. I'm not sure what he's using the annotation wrapper for, but I'd check on the validity and function first.
> 
> ```
> 	[WorldForge]
> 	[CommandProperty(AccessLevel.GameMaster)]
> 	public int MaxRange => 0;
> ```

Good point on the annotation.

PR #1467, opened 2022-03-20
Need more help on thaum summon. This again is for testing. If too pow…

Chilly021:
…erful, which I assume it might be, it will be scaled back

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Phantasm/SummonedDjinn.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Phantasm/SummonedDjinn.cs
@@ -14,10 +14,11 @@ namespace Kesmai.Server.Game
 		{
 			Summoned = true;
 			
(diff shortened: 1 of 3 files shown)


Wirednuke:
My name is Brick, and I approve this message.

dedmist:
Is there a way to get the caster info during construction?  Ideally they would scale hp/block with magic level.

SewerCzar (approved):

SewerCzar:
> Is there a way to get the caster info during construction? Ideally they would scale hp/block with magic level.

Scaling 100% needs to happen. I think it's on Lux's huge list of stuff to do.

For now, 1200 defense on a level 9 spell (phantoms are 9, right?), 1250 on an 11, and 1500 on a 17 spell are solid. Especially since they can't balm or really defend themselves. Worst thing that happens is thaums are SLIGHTLY overpowered for a while and that's fine.

dedmist (approved):
Fair enough

Leolus (approved):

PR #1475, opened 2022-03-22
Thief Trainer in Lockpick Town

Wirednuke:
Added a Thief Trainer to lockpick town, giving thieves a training option that allows them the ability to push gold from a bank. 

Created a Thief area behind a secret door to the north of the wizard trainer in lockpick. 

Second Commit added a teleporter to keep the riff raff out of the thieves guild.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -47846,6 +47846,123 @@
       <name>Lockpick Town</name>
       <height>150</height>
       <level>14</level>
+      <tile x="6" y="-2">
+        <component type="WallComponent">
+          <wall>149</wall>
+          <destroyed>150</destroyed>
+          <ruins>0</ruins>
+          <indestructible>true</indestructible>
+        </component>
+      </tile>
+      <tile x="7" y="-2">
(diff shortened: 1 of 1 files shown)


dedmist (approved):
I don't see a problem with this.

Looks like you forgot to add a barrel or two along the path though.

Chilly021:
Probably need to do something about the other trainer? Just figure this makes the other obsolete being so close.

Tawnos-LOK (approved):
I think Lockpick Town thematically makes sense for this so I'm appoving.  Some suggestions for the future - maybe add another door by Lamont (banker) to connect to the back hallway to the east and add one barrel along the path to avoid.

Also, to Chilly's point, something will have to be done with the preexisting trainer down below by the rope.  Possibly just removed / possibly made into a vendor (thief only?).

jkachhad:
Solution coming to gold weight. I would try to say rather than band-aid a fixes in (banker vendors), is there a game-play feature we can create for players to invest into?  I’ll implement it this weekend :)

jkachhad:
New rings added to add some meta to hunting and gold carrying.

PR #1476, opened 2022-03-22
Many of the level 8 crits in Axe had lower exp or same exp as level 6…

Chilly021:
… crits. Bringing these to be 3-3.9k range

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -104228,7 +104228,7 @@
         MaxHealth = 57, Health = 57,
         BaseDodge = 15,
         HideDetection = 23,        
-        Experience = 1881,
+        Experience = 3881,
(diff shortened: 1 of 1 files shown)


SewerCzar (commented):
The only level 6 crit that has more xp than a level 8 anything is the gargoyle. The level 8 goblins are 500 more than level 6 goblins. In fact this boosts those crits over the harder mobs in the zones. 

Please reach out if values need changing but changing individual crits is going to lead to the system being like it was previously

dedmist:
The standardization changes are made to handle situations of determining scaling, that way we won't have questionable zoning xp and difficulty scaling; if we feel that the xp is underleveled in an area we would instead boost the crit level.
The current areas (except in kes) already follow a metric unless they were changed somewhere along the line.

PR #1478, opened 2022-03-23
Throwable preference on Main Gauche

Wirednuke:
Ok, I need a little help from the big boys on this one. I want to take the throwable flag off the Main Gauche so it just sits in your offhand like a regular weapon (it's really too big of a dagger to throw anyway) and allow the thief to press / and throw his rdagger from the main hand without it launching the blocking weapon.

So what I did:

- Removed base.flags from MainGauche.cs and inserted the weapon flags I wanted to keep from the dagger base class. This should remove the throwable/quickthrow flags from the base dagger class?

@jkachhad  Does this work the way I think it does? Appreciate some help/input, or even better.. .if you know a way to fix this issue that doesn't require us to remove the throw flags.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -100072,8 +100072,9 @@
         new CreatureBasicAttack(18, 18, 30)
     };
     
-    carfel.Wield(new MainGauche());
     carfel.Wield(new YasnakiDagger());
+    carfel.Wield(new MainGauche());
+    
         
     carfel.AddGold(4800);
     carfel.AddLoot(new LootPack(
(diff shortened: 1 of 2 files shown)


Wirednuke:
Also tried to switch which hand carfel is holding the dagger in. He's evil and the dagger is neutral, so he's fumbling it. Figured maybe if I listed it second it would be in his offhand instead. 

If this doesn't make it better I'll take neutral off the dagger.

Tawnos-LOK (approved):

yarrim (approved):

zrikzlok (approved):
Well see if it works on next restart :p

PR #1480, opened 2022-03-23
Currently Yas Ninjas have a 20 skill attack, which is 3 higher than shid …

Chilly021:
…ninjas and then way higher than most crits/mobs and higher than most lair crits (carfel). Also they have a Medium Penetration, which Yas has Heavy. I still disagree with a hand having heavy pen unless they have some special gauntlets they wear/drop. TK Elite Guards are quite a bit tougher now due to Halberds were fixed. I don't believe just because they are Elite Guards, they should have a Heavy Pen pen. It should be based on Weapon. Additionally due to the crit/mob density and such they represent way more of a threat now than TK himself. There are 11 guards now with Heavy Pen and level 18 skill. I am moving it back to 16. I am letting the pen inherit from base of halberd now (Medium) and the skill the be more representative of a level14 crit

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -100854,9 +100854,6 @@
         Experience = 10699,
         
         CanCharge = true,
-        
-        BasePenetration = ShieldPenetration.Medium,
-    
     };
     
     
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -100851,18 +100851,18 @@
         MaxHealth = 209, Health = 209,
         BaseDodge = 27,
         HideDetection = 29,    
-        Experience = 11000,
+        Experience = 10699,
         
         CanCharge = true,
         
-        BasePenetration = ShieldPenetration.Heavy,
+        BasePenetration = ShieldPenetration.Medium,
     
     };
     
     
     trollguard.Attacks = new CreatureAttackCollection()
     {
-        new CreatureBasicAttack(18)
+        new CreatureBasicAttack(16)
     };
     
     trollguard.Wield(new Halberd());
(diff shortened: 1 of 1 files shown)


Chilly021:
Shid Ninjas in comparison: 

` ninja.Attacks = new CreatureAttackCollection
    {
        {
            new CreatureAttack(17,      10, 18,      "The ninja hits you with a flurry of fists.",
                                                        new AttackProneComponent(5), 
                                                        new AttackStunComponent(5)),                40
        },
        {
            new CreatureAttack(17,      13, 20,     "The ninja slams his fist into your chest.",
                                                        new AttackProneComponent(5), 
                                                        new AttackStunComponent(5)),               40
        },
        {
            new CreatureAttack(17,      15, 21,     "The ninja knocks you back with a forceful kick.",
                                                        new AttackProneComponent(5), 
                                                        new AttackStunComponent(5)),               20
        },
    };`

sproatba:
I'm not sure the change to ninjas is needed, non twink 16/6th dan was a pretty easy solo.  I only had 1 dex and 2 med shield so I wasn't even maxed out on defensive items yet.    I didn't even pull the ninjas correctly 1 group at a time and was able to tank 2 groups just fine even though it did use alot of balm.  Did something else change recently to prompt this?  With kosh being about 14/4th for kosh gaunts i'm not sure we want shid to be too fast after that unless we have other changes in mind.

Chilly021:
> I'm not sure the change to ninjas is needed, non twink 16/6th dan was a pretty easy solo. I only had 1 dex and 2 med shield so I wasn't even maxed out on defensive items yet. I didn't even pull the ninjas correctly 1 group at a time and was able to tank 2 groups just fine even though it did use alot of balm. Did something else change recently to prompt this? With kosh being about 14/4th for kosh gaunts i'm not sure we want shid to be too fast after that unless we have other changes in mind.

It was easy to solo because Halberds were bugged giving a minus to their AR. This was fixed. If you go there now, you will see the change.

dedmist (approved):
After the halberd change, I also noticed that the guards seem to block quite a bit more than TK.  Would be nice to verify if this is halberd or rng.

against 13 rdag

Chilly021:
Also the ninjas for Shid's lair are not changing, only the ninjas for Yas. They are only going to 18, where-as Shid's are 17. Since Yas are is not really deemed a "lair" and there are Curse thaums and other gotchas there, it seems tad bit overkill they are the only crits that have that high in weapon skill + Heavy penetration. We can add back the heavy pen if needed, but I am not sure why that was added other than it was the only was to know for sure the player block percentage would be totally hosed.

sproatba:
i'm good with halberd change, i'm confused on where the ninja nerf are coming from.

SewerCzar (approved):
Yas ninjas should actually be 17 attack value - was editing a lot of mobs it slipped through.

Troll guards are actually kind of weak right now imo. Their attack value is actually up too. So I approve but might need to look at it a bit more

Chilly021:
You will need to go up and test on those Guards then. I am only tweaking on feed back and now first hand. They hit SUPER often and 4+ at a time can be a problem to even a 8th/9th dan.

Wirednuke:
> Yas ninjas should actually be 17 attack value - was editing a lot of mobs it slipped through.
> 
> Troll guards are actually kind of weak right now imo. Their attack value is actually up too. So I approve but might need to look at it a bit more

I haven't had an issue with their attack on my thief... their DR is insane though. They are at like 1500, and I can barely hit them at 13 dagger unless it's enchanted. On the bright side, venom takes care of them nicely when you finally do land something lol.

Tawnos-LOK (approved):

PR #1490, opened 2022-03-24
Remove Poison Cloud Immunity on all. Keep Poison Immunity on Draugr

Chilly021:
- Poison Cloud will not hurt Draugrs, only blind them
- All other poison cloud immunity has been removed from non draugr


SewerCzar:
I'm for this for now but long run I think the solution has to be turn undead being effective because poison affecting undead is weird

SewerCzar (approved):

yarrim (approved):

Leolus (approved):

PR #1491, opened 2022-03-25
UK Gargoyle converted to lurker

SewerCzar:
Converting the gargoyle to a more appropriate monster, and toning it down to one that spawns every 7-10 minutes instead of 3.5 - 5.

Has high attack value, has poison, but doesn't respawn like crazy and punish people trying to hunt in the zone.

Between the pillar and a single monster this will adequately stop scripting without going overboard like it is now according to feedback.


Chilly021:
This is super agreeable. The spawn time of Gargs make Prin Runs really not doable unless you have a group at this time. The spawn time is a bit too much and a lurker will probably best to keep people honest from scripting.

Chilly021 (approved):

zrikzlok (approved):

dedmist (approved):
Yup, good call

PR #1492, opened 2022-03-25
Changes to what mobs flee

SewerCzar:
Trolls over level 5 and Orcs over level 10 no longer can flee. Goblins up to level 15 have the canflee flag.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -104469,6 +104469,7 @@
         BaseDodge = 13,
         HideDetection = 22,        
         Experience = 1306,
+        CanFlee = true,
     };
     
     goblin.Attacks = new CreatureAttackCollection()
@@ -104566,8 +104567,6 @@
         Experience = 1804,
                 
         Movement = 2,
-                
-        CanFlee = true,
     };
     
     troll.Attacks = new CreatureAttackCollection()
@@ -104615,6 +104614,7 @@
         BaseDodge = 13,
         HideDetection = 22,         
         Experience = 1306,
+        CanFlee = true,
     };
     
     goblin.Attacks = new CreatureAttackCollection()
@@ -104724,6 +104724,7 @@
         BaseDodge = 14,
         HideDetection = 22,         
         Experience = 1306,
+        CanFlee = true,
     };
     
     goblin.Attacks = new CreatureAttackCollection()
@@ -104781,6 +104782,7 @@
         HideDetection = 23,        
         Experience = 1881,
         BasePenetration = ShieldPenetration.VeryLight,
+        CanFlee = true,
     };
     
     goblin.Attacks = new CreatureAttackCollection()
@@ -104829,8 +104831,6 @@
         Experience = 2700,
         BasePenetration = ShieldPenetration.Light,
         Movement = 2,
-                
-        CanFlee = true,
     };
     
     troll.Attacks = new CreatureAttackCollection()
@@ -105218,6 +105218,7 @@
         HideDetection = 23,        
         Experience = 1881,
         BasePenetration = ShieldPenetration.VeryLight,
+        CanFlee = true,
     };
     
     goblin.Attacks = new CreatureAttackCollection()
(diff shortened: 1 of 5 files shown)


sproatba:
Asheron has a whole thing in the works to simplify this type of stuff, might be worth fleshing that out instead of bulk changes like this at this point.   Also, historically the game wasn't like this, different stuff on different levels would run away and you just kind of had to learn what happened where.  Trolls on TT always ran, i'm sure they are higher than lvl 5.  Level 5 is pretty arbitrary to anyone game side without access to the definition of what level 5 is so it will seem just about as haphazard as it does now unless something is also done to indicate a "level up"  Troll, Elder Troll, Elite Troll, or something of that nature.  right now we distinguish mobs by adjectives describing weapons or spells they use because there is no sense of level or name distinction. (lhob, speedy hob, hob, ma gob)  I think the "level up" system is far better than random and arbitrary, but without something a little more, we either lose some historical accuracy or have to break the system.  We also very quickly will run out of running mobs if they all eventually quit running once they level up high enough.   What runs past 15?  Do we just quit having runners?

dedmist:
I think the consistency of flee is a little weird, but I prefer the mechanics as it adds variety even if it's sometimes frustrating.  I would like to see if other people feel like fleeing is a burden or a game mechanic.

SewerCzar:
> I think the consistency of flee is a little weird, but I prefer the mechanics as it adds variety even if it's sometimes frustrating. I would like to see if other people feel like fleeing is a burden or a game mechanic.

I think you'd be hard pressed to find anyone who actually likes flee mechanic haha. But I'd like it to make sense in terms of which mobs can/will do it. Like a player enters a zone sees goblin can expect it to possibly flee. Trolls and orcs are kind of random where they are as is.

Chilly021:
Honestly, I'm of the mindset flee should be removed from the game.

SewerCzar:
> Honestly, I'm of the mindset flee should be removed from the game.

I mean that's fine by me. (maybe leave the mechanic in for monsters like Greedy Goblin). If you want to make a pull popping it off all current mobs for Lux to try I'll approve it. But since it's in I want it to make sense, hence the pull.

Chilly021:
I am just trying to make sense of it. I feel it was an artifact of the old AOL model of pay by the hour. The mechanic doesn't make since due to the game has been dramatically sped up with faster facets. I would be willing to argue otherwise. I mean greedy gobs and maybe casters make sense, but melee crits running off is just annoying. I don't see it adding to difficulty in fact taking away from it. If you are gullible to try to hunt the crit down.. then you are putting yourself at risk in higher areas.

jkachhad:
```
var hitsAsPercent = (Health * 100 / MaxHealth);

			if (hitsAsPercent < 10 && CanFlee)
				Flee();
				```
				
A creature's ability to flee is determine by their health.  Anything with less than 10% of their health and CanFlee = true, will run.
I think all `Humanoids` should flee. 

It's all about choices.  If you know you may be up against many humanoids, you might opt to carry a returning weapon to catch them.  Those without access are at a disadvantage.   if nothing can flee, you might as well put on a shield and keep slamming the * key.  Any tanking class would get a benefit.  There is no reason for MAs to consider Jumpkicking, other than to catch caster fizzle or ranged attackers.

SewerCzar:
> ```
> var hitsAsPercent = (Health * 100 / MaxHealth);
> 
> 			if (hitsAsPercent < 10 && CanFlee)
> 				Flee();
> ```
> 
> A creature's ability to flee is determine by their health. Anything with less than 10% of their health and CanFlee = true, will run. I think all `Humanoids` should flee.
> 
> It's all about choices. If you know you may be up against many humanoids, you might opt to carry a returning weapon to catch them. Those without access are at a disadvantage. if nothing can flee, you might as well put on a shield and keep slamming the * key. Any tanking class would get a benefit. There is no reason for MAs to consider Jumpkicking, other than to catch caster fizzle or ranged attackers.

Want us to put CanFlee = true as a base class in humanoids? Would be the easiest solution to what you want

PR #1497, opened 2022-03-26
ITreasureUpdate

sproatba:
Added ITreasure to several items players wouldn't want to go poof when disintegrate is cast.  If I missed any or we disagree with any settings let me know and I'll update.  There is an issue out there for the moonstone ring that will need closed as well if this goes.


Tawnos-LOK:
Some people might revolt if napthas disintegrate, as was changed in one of the commits here.

SewerCzar (commented):
I'd say no to naptha bottles. Monsters running around with sacks full of them after people hunted a zone wouldn't be right. Conc bottles would need same treatment.

Edit: Whoops naptha is being changed to no longer disintegrate. 

Can't you buy a new spellbook for cheap? Do we want monsters picking them up and running with it?

Rest seem good though. Edit: Spellbook isn't a problem I was just wondering about it.

SewerCzar (approved):

sproatba:
I could go either way on spell book, once you get past early levels most don't carry it anyways.

sproatba:
I'd say if naptha stays, so should conc, but not a big deal either way to me.

SewerCzar:
> I'd say if naptha stays, so should conc, but not a big deal either way to me.

No it was just me mis-reading it. Bottles shouldn't be un-disintegratable. I'm good with the pull. I don't like non treasure taking up monster carry slots. Maybe coal rocks should be taken off treasure too? Dunno but this pull is fine I approved it.

jkachhad:
```
public virtual bool CanDisintegrate => (this is not ITreasure) && (Quality < ItemQuality.Uncommon);
```

Updated internally.

Chilly021:
Just in case, there is another PR that has a lot of Base level changes as well. Should we take out two birds with one stone? Just trying to make sure you don't have any merge conflicts here since this is doing more than one thing illustrated in the job/feature.

sproatba:
> Just in case, there is another PR that has a lot of Base level changes as well. Should we take out two birds with one stone? Just trying to make sure you don't have any merge conflicts here since this is doing more than one thing illustrated in the job/feature.
> 
> https://github.com/jkachhad/Stormhalter/pull/1434/files

A quick look through that one, and I doubt we'll have any conflicts between these two.  It should be smart enough to merge them on it's own.  we can catch either branch up with master and resubmit once one is passed as well without too much difficulty, just need to rebase.

Chilly021 (approved):

Chilly021:
@Tawnos-LOK  @jkachhad  @zrikzlok @yarrim  @Leolus @dedmist  This looks good to me. Would be nice to make sure stuff doesn't get disintegrated that shouldn't. - Bumping this for approval.

zrikzlok (approved):

PR #1504, opened 2022-03-28
Feature/movie fun bosses

Chilly021:
* Making use of some empty space in Leng
* Add a small tribute to a great movie. 
* Walls Indestructible
* One Entrance, Two Exits
* Three Mini Bosses: Jim The Enchanter, The Black Knight, and the Killer Bunny
* Enchanter can not move. 
* No New Items at this time
* This is a test for hopeful new Weapon/Item drop

image

image

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -78879,9 +78879,10 @@
 ]]></block>
         <block><![CDATA[]]></block>
       </script>
-      <script name="OnDeath" enabled="false">
+      <script name="OnDeath" enabled="true">
         <block><![CDATA[]]></block>
         <block><![CDATA[
+    source.SayTo(killer, $"Only a flesh wound! {killer.Name}.");
 ]]></block>
         <block><![CDATA[]]></block>
       </script>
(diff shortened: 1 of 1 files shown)


Chilly021:
image

yarrim:
The door is not very secret. intentional?
Also, perhaps the Black.Knight should roam? I think at the moment he will pursue, but stand still if no targets are in sight, leaving him vulnerable to a thief with a doom bow.

Chilly021:
Good Feedback! I'll see about some AI here to take care of that. He will pursue, but he's not meant to be entirely difficult since it's more of a scenario based set of bosses to encourage more fun and throughput. Also he's not meant to be plinked at either so hide detection should be pretty high. Any 

Additionally, for the door, correct, not secret at all. More of a mechanic to hold the Rabbit at bay. If someone leaves the door open per se, it defeats the purpose obviously. Best peak the rabbit to ensure that doesn't happen. 

Two additional Notes:

* Yes the enchanter can be out ranged and has 0 movement. This is purposeful, should you get too close without actually interacting with the Black Knight. You can still pull the Black Knight off screen to fight him. 
* If a user does not have enough balms and/or they wish to only solo the first two, without trying the Killer.Rabbit then they can step out of the Lower SE blood hex.

SewerCzar (approved):
Haha. I like it.

yarrim (approved):

zrikzlok (approved):

PR #1507, opened 2022-03-28
Bug fix for Gem Price Mutators and correct comments

TimPickup:
Exactly the same fix on all appropriate maps.

gemsPriceMutatorMedium was intended to be a modifier of -40% to +nothing. Instead it was doing -40% to + 100%

gemsPriceMutatorAboveAverage was intended to be a modifier of + nothing to +40%. Instead it was doing +100% to +40%

Also update the comments on every mutator to reflect the correct values

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -106,9 +106,9 @@
         item.Price = Utility.RandomBetween(-lesserPrice, higherPrice);
     }
     
-        public static void gemsPriceMutatorLow(MobileEntity entity, Container source, ItemEntity item)
+    public static void gemsPriceMutatorLow(MobileEntity entity, Container source, ItemEntity item)
     {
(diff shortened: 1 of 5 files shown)


sproatba:
looks good, but I have a feeling we are going to start seeing a lot of crappy gem value comments when this goes in.

TimPickup:
> looks good, but I have a feeling we are going to start seeing a lot of crappy gem value comments when this goes in.

Its only a small fix. If you want to increase the values then do so, but currently it doesnt do what it was intended to do, and currently the same gem from a low level creature has a good chance to be worth More than from a Dragon or Drake etc!

SewerCzar (approved):
I dont know how we put the same pull in simultaneously but thanks. As for gems being worse off for this: It's actually going to raise the prices of a lot of gems. Troll gems specifically.

yarrim (approved):

Leolus (approved):

PR #1508, opened 2022-03-28
Fixed issue for gems

SewerCzar:
Medium gems will now cap at high end of gem base value, Above average at low end of gem base value. Descriptors in the file are now accurate.

If an int of 0 doesn't work please comment. Other than that this fixes some gem issues that have been noticed


sproatba:
[there is already another pull out there that is pretty much identical to this](https://github.com/jkachhad/Stormhalter/pull/1507)

PR #1513, opened 2022-03-28
Upgrade Dungeon of Doom

TimPickup:
Dungeon of Doom is only accessible on route between Ogre Caves (a level 10 area) and Undead (a level 16 area).

Currently the area was spawning Level 8 creatures. I believe an oversight due to it descending where all other progression in Axe ascends.

I have made Level 13 version of most creatures on this level by copying Level 10 version  and then applying similar exp and health etc values from the same or similar creatures of level 13 in Leng.

I know its a fairly big update and will need testing but currently the area makes no sense as its regressing before a massive jump up. I believe this should be in keeping with the area while providing progression between the areas

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Axe Glacier" version="0.92.0.0">
+<segment name="Axe Glacier" version="0.94.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

Chilly021:
Makes sense, wish we had lvl 12 crits, but the math looks correct here.

TimPickup:
> Makes sense, wish we had lvl 12 crits, but the math looks correct here. As mentioned there are still lvl 10 crits still there though. If anyone has any ideas to remove them.

I would be happy to make lvl13 Bear, MiniBossBear, and RockWorm if people want that too. I only left them as is to not make the area quite so hard but makes sense to upgrade them too.

As I mentioned I put the archers to lvl 10 because previously they were lvl 6 (less than all the other creatures) so I presumed the area wasnt intended to be hard vs archers originally. I did make a lvl 13 archer though so happy to use them if people prefer

I also don't think Prince Wolf really has a place here - no other wolves and hes outclassed by other creatures.

sproatba:
seems like it's on the right track.  I'd almost say dump the lvl 10 stuff, but that would get rid of all the stuff requires sga.

SewerCzar:
> 

I like what you're doing here. It makes a lot of sense

I've added a file that drops the wildlife from the dungeon of doom (I feel that it doesn't make too much sense for wolves and bears to be hanging there - kwolf can still spawn in it though) and has added level 13 version of rockworm. Stats for a level 13 bear can be done up if you want to add it. 

https://docs.google.com/spreadsheets/d/1WE_r9vK09VXieHkj5PwqRPswFuQR66oafhSTdFpP4Q8/edit?usp=sharing

https://docs.google.com/spreadsheets/d/1D9tmGWRpw2KKr4NuMhjo8Owr3GUvnTYN0HXXfzwrWUM/edit?usp=sharing

https://docs.google.com/spreadsheets/d/1D-COAB-Bd2FhfUY2czKc8kbyjeBMpzhj4fFMad_q-jE/edit?usp=sharing

https://docs.google.com/spreadsheets/d/1B30CgcseVxJgdKiCCiCMEB9itHtqGSNuj7pXxwOhRak/edit?usp=sharing

^^ Those are all the numbers that were used. They're by no means set in stone they're just the current system in place


[dungeonofdoom.zip](https://github.com/jkachhad/Stormhalter/files/8372302/dungeonofdoom.zip)

Leolus (approved):

Tawnos-LOK (approved):

tmesolutions:
> > 
> 
> I like what you're doing here. It makes a lot of sense
> 
> I've added a file that drops the wildlife from the dungeon of doom (I feel that it doesn't make too much sense for wolves and bears to be hanging there - kwolf can still spawn in it though) and has added level 13 version of rockworm. Stats for a level 13 bear can be done up if you want to add it.
> 
> https://docs.google.com/spreadsheets/d/1WE_r9vK09VXieHkj5PwqRPswFuQR66oafhSTdFpP4Q8/edit?usp=sharing
> 
> https://docs.google.com/spreadsheets/d/1D9tmGWRpw2KKr4NuMhjo8Owr3GUvnTYN0HXXfzwrWUM/edit?usp=sharing
> 
> https://docs.google.com/spreadsheets/d/1D-COAB-Bd2FhfUY2czKc8kbyjeBMpzhj4fFMad_q-jE/edit?usp=sharing
> 
> https://docs.google.com/spreadsheets/d/1B30CgcseVxJgdKiCCiCMEB9itHtqGSNuj7pXxwOhRak/edit?usp=sharing
> 
> ^^ Those are all the numbers that were used. They're by no means set in stone they're just the current system in place
> 
> [dungeonofdoom.zip](https://github.com/jkachhad/Stormhalter/files/8372302/dungeonofdoom.zip)

Thank you @SewerCzar for those. Very useful for further changes. Good to have a definitive guide.

Really glad this got approved. I was worried it would be waisted time but I'm glad we can see how it goes and develop it further.  I'm happy my small offering is of use.

PR #1516, opened 2022-03-29
Enhancements to the Movie Scenario

Chilly021:
* All Crits are Evil now (So no cheesing)
* Added Death Dialog to the Enchanter
* Add Gold at least to bunny rabbit since there is no real value for doing the scenario other than fun.
* Add Night Vision Detection
* Add Hide Level 45 - These crits should be fine doing them one at a time. Also you should be able to leave if you can not finish to the east (blood hexes)

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -78842,10 +78842,11 @@
     {
         Name = "Black.Knight",
         Body = 193,
+        Alignment = Alignment.Evil,
         
         MaxHealth = 800, Health = 800,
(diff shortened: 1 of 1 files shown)


SewerCzar (commented):
Hide level 45 requires a 26 thief with 19 magic seems a bit overkill..

SewerCzar (approved):

Leolus (approved):

Tawnos-LOK (approved):

Issue #1503, opened 2022-03-27
Unexplained spawn placement


yarrim:
Angry found a mob trapped behind the counter in the Giant castle.
This tile and the counter should be exclusions for the region spawner, so I'm not sure how the mob got there. Either entities can wander into that tile (the walls and counters all look right to me) or it spawned there despite being an exclusion.

trapped npc:
![image](https://user-images.githubusercontent.com/11442233/160306703-fa5dc71e-24a3-47c1-b568-8364650987b4.png)

spawner:
![image](https://user-images.githubusercontent.com/11442233/160306693-ec92d52d-29d9-420c-8031-fab7fbf058a1.png)

yarrim:
another reported location is Giant-Exterior mobs showing up inside the castle. Exclusions being off-by-one would explain this:
![image](https://user-images.githubusercontent.com/11442233/160307870-25518437-5b5d-4e95-8a0d-2690315bf9f0.png)

SewerCzar:
https://github.com/jkachhad/Stormhalter/pull/1518

I should have fixed the specific mentions here besides the mystery NPC which is odd as I also see an exclusion. I've also fixed that one 3x3 square and the line missing in the spawn outside. 

Should be good to approve please check it out

Chilly021:
Fixed in : https://github.com/jkachhad/Stormhalter/pull/1558

PR #1518, opened 2022-03-29
Giants Castle Edits

SewerCzar:
https://github.com/jkachhad/Stormhalter/issues/1503

As noted here I've done some exclusions inside the giants castle, and included some edits outside that will no longer cut the "Outside Giants Castle" spawn in half. Spawn should be cleaner now


Chilly021 (approved):

Chilly021 (commented):
Actually need a few more eyes on this. I don't have time to check this out today.

Chilly021 (changes requested):
@sproatba  Can you proof read these, I know you were looking into this.

sproatba:
yeah, just trying to spell out where it stood so others could process easily.

PR #1519, opened 2022-03-30
Axe edits

SewerCzar:
Renamed the spawns so that theyre organized relatively together.

Added night vision to curse trolls as there's very little threat if using darkness right now. Almost negligible.

If curse troll is deemed too much we can do two things:

Decrease number of curse trolls.

Should fix spectral mage spells finally. Ice storm should be fine damage wise to players (same as yeti right now which isn't much). Might need tuning

Drop curse trolls night vision and increase number of wargs,

Other zones need similar treatment after playing wizard and thief.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -111341,7 +111341,7 @@
     <spawn type="RegionSpawner" name="Undead Northeast">
       <minimumDelay>900</minimumDelay>
       <maximumDelay>1500</maximumDelay>
-      <maximum>49</maximum>
+      <maximum>45</maximum>
       <script name="OnAfterSpawn" enabled="false">
         <block><![CDATA[]]></block>
         <block><![CDATA[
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

Leolus (approved):
Please add ice fix attempt to notes and I'm good with this.

sproatba:
image
not sure what intensity is here, what did it do?

I've not spent enough time down there to really comment, but it just looks like reducing the number of overall mobs spawned and reducing the percentage of trolls along with a bunch of spawner renames which make it easier to group spawners.  Spawner rename is def good, mob density in axe seems over the top overall so i'm guessing the reduction is warranted.

Chilly021 (approved):

Chilly021:
I'm leaning on the others here. Agreed on Icestorm.

SewerCzar:
Addressing all comments: Removing the intensity should actually make the spectres cast a fairly neutral high ice storm that won't do much damage with f/i on.

Trolls will have night vision so players will have to look out for some slightly enhanced danger if using darkness (Was playing there with Vothric and in my opinion it's needed)

Density is going down only slightly and it's the trolls. I do think we probably need a big overhaul on axe/leng density including axe UD but people will really not be happy. I think UK and Oak are the only ones based in OG logic, and the new ones are... insane. It's my fault from back in July when I didn't know enough.

zrikzlok (approved):

PR #1547, opened 2022-04-04
Add Ram Skull Staff

Chilly021:


Source/Kesmai.Server/Game/Items/Equipment/Weapons/Staves/RamStaff.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <inheritdoc />
+        public override int ProjectileProtection  => 1;
+
+		/// <inheritdoc />
+		public override WeaponFlags Flags => WeaponFlags.TwoHanded | WeaponFlags.Bashing | WeaponFlags.Lawful
+		
+		/// <summary>
+		/// Initializes a new instance of the <see cref="BlackStaff"/> class.
+		/// </summary>
+		public BlackStaff() : base(901)
```
sproatba, line 49: Public RamStaff() here

sproatba:
4 base armor is probably too strong if this is going to be in leng area you are working on

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Staves/RamStaff.cs
Chilly021: oof good catch - removed that and added.

Chilly021:
> 

We are hearing a huge area of complaints in regards to Staves in general. Since they are 2H, you can not typically wield a shield without depreciating the DR currently. I can make this 1H but it creates ambiguity. This should be an intermediate blocker until OL. If staves do not have block currently, they are functionally useless to casters.

Wirednuke:
> 4 base armor is probably too strong if this is going to be in leng area you are working on

Is there a way to give it a flat magic damage reduction? like maybe 10-20%? To me, melee damage has never really been a problem with my casters... it's been getting trashed by magic casters.

Making it 3 Base armor Bonus would make it comparable to the mace/shield combination...

Wirednuke:
> > > 4 base armor is probably too strong if this is going to be in leng area you are working on
> > 
> > 
> > Is there a way to give it a flat magic damage reduction? like maybe 10-20%? To me, melee damage has never really been a problem with my casters... it's been getting trashed by magic casters.
> > Making it 3 Base armor Bonus would make it comparable to the mace/shield combination...
> 
> I see Mace and Shield a even trade off due to it is Ranged for +3. Staves have no ranged, so I am bumping this to a 4.

Fair enough. Obviously this will help make casters stronger, but I don't think more armor fixes their squishy problem. I think they're going to keep geting wrecked due to lack of hp/magic mitigation. 

I think staff is just what people latch onto because they see the other weapon changes.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Staves/RamStaff.cs
Tenscore: #1545 hasn't gone through yet, and changes your ID to 902
(To keep the non poison and poison daggers together.)

sproatba:
Not entirely sure what do do about squishy casters but giving them the best blocking weapon in the game to offset lack of adds/hps just forces weapon choices.  Perhaps we need some more range on blocking to give weapons more flavor, but +4 on staff is the difference between leather and ices, that's alot for a bg weapon.  This is probably up there for one of the best weapons in the game as it stands outside of lairs where it can't hit.

Chilly021:
I would say if we do put this down to 3, we put the Fan back down to 3 as well.

Chilly021:
This also matched the new Dagger MainGauche: 

	public override int BaseArmorBonus => 4;
		
		/// <inheritdoc />
		public override int BaseAttackBonus => 2;

Wirednuke:
> I'm not disagreeing, but I do not see the needs being met from a class perspective. I am willing to bump this down to a 3, but we need a follow up item to fix resistances on classes. You don't have to hunt in a heavily caster area on a Wiz/Thaum otherwise until then. This is still a wiffle ball bat compared to other weapons on the AR / 1-8 dmg and all that. There is not a blue glow on this or silver either. This is comparable to the Fan without projectile of 4.
> 
> ```
> 	public override int BaseArmorBonus => 4;
> 
> 	/// <inherit />
> 	public override int ProjectileProtection => 4;
> ```
> 
> Here are the other comparisons:
> 
> 2-H Knight Hally:
> 
> ```
> public override int MinimumDamage => 2;
> 
> 	/// <inheritdoc />
> 	public override int MaximumDamage => 14;
> 
> 	/// <inheritdoc />
> 	public override int BaseArmorBonus => 2;
> 
> 	/// <inheritdoc />
> 	public override int BaseAttackBonus => 5;
> ```
> 
> 1 H RAxe:
> 
> ```
> public override int MinimumDamage => 1;
> 
> 	/// <inheritdoc />
> 	public override int MaximumDamage => 8;
> 
> 	/// <inheritdoc />
> 	public override int BaseArmorBonus => 2;
> 
> 	/// <inheritdoc />
> 	public override int BaseAttackBonus => 5;
> ```
> 
> SGA:
> 
> ublic override int MinimumDamage => 4;
> 
> ```
> 	/// <inheritdoc />
> 	public override int MaximumDamage => 14;
> 
> 	/// <inheritdoc />
> 	public override int BaseArmorBonus => 1;
> 
> 	/// <inheritdoc />
> 	public override int BaseAttackBonus => 4;
> ```
> 
> BBS:
> 
> public override int MinimumDamage => 2;
> 
> ```
> 	/// <inheritdoc />
> 	public override int MaximumDamage => 12;
> 
> 	/// <inheritdoc />
> 	public override int BaseArmorBonus => 3;
> 
> 	/// <inheritdoc />
> 	public override int BaseAttackBonus => 4;
> ```

Fan and MainGauche are neutral which keeps knights (those most likely to exploit it) from using it without giving up their class perks. I'm not against revisiting their blocking bonus if the community feels that way, but that's a separate issue anyway.

My issue with the staff, wasn't that the 4 armor bonus was too much... it's that I don't really think it's going to fix the problem. If anything, it'll give casters a false sense of confidence that'll lead them to hunt in areas where the magic and ranged damage is even worse. But put it in, and lets see what happens.

sproatba:
yeah, so far my experience has been you quickly are at or near block cap, but never get anything to deal with casters, or casters and archers for ma.  With the lack of hps this punishes casters the most.  Most areas melee are easy, you can gather up several groups and deal with them once you get going.  You only really have a few ways to get stronger, skill, items, and level.  casters get hit hard on level, and items outclass skill by quite a bit for combat.  They are really lacking in the magic department.  Resistances don't resist enough at what feels level appropriate because of the melee being outclassed by items.  I mean -4 with blind res and 16+ will power you still get blinded 95% of the time at levels where you can't even train in kes.  Unfortunately I just don't think we have the tuning options available to fix this stuff until spells get some love.  we can run around nerfing all the casters, but that is just a Band-Aid.

Wirednuke:
> So all in agreement with +4 armor and getting class needs worked on separately? I think there is some derailment here. We don't have access to the back end code. So let's get the quick wins for now. This staff could be turned into a cool "auto" shield enhancement too or + to Fire spells or whatever later on.
> 
> If you have not hunted solo in Bloodlands, -280 or Axe UD with a staff, then that will give you some idea. Melee destroys you since most have NV and higher level.
> 
> Yes resistances need to be worked on, but let's get that issue nipped in the bud somewhere else, this is just to get a "blocking" option 2H in play.
> 
> It's still a wiffle ball bat until the new weapon properties are added.

Can't help you with approvals, but I concur. See Issue #1548

SewerCzar (approved):

Leolus (approved):

Wirednuke (approved):

zrikzlok (approved):

PR #1550, opened 2022-04-04
Swapping griffin and Dragon Skull stats

SewerCzar:
We've talked about this change for a while and it makes sense item progression wise

Full disclosure: I do have a couple dragon skulls in the bank


dedmist:
The f/I on Griff is pretty much a requirement for kosh its a good mechanic.

SewerCzar:
> The f/I on Griff is pretty much a requirement for kosh its a good mechanic.

Yeah a full on swap isn't right but griff skull needs to be limited. Maybe daisy drops a fire protection no night vision helm then gryphon has night vision and stun protect? This isn't it though

PR #1551, opened 2022-04-04
Mana/Cast ration balance

SewerCzar:
Noted by players that some casters had extreme amounts of casting occuring:

Kes -> UK Decreased the amount of magic missiles cast across the board (average monster has the chance of 2-3 spells before coming in on the player - will still exist to cast more)

UK Yasnaki Thaumaturges are decreased from 7 curse to 4 curse - 3 monsters in a pack vs 4 undead banshees that are 2.7 and will be nerfed soon. They will also cast 2 sets before closing in one players.

UK Undead: Spectre will now cast a maximum of 3 deaths instead of potentially 6 from before.

Oak Undead, the spectres will cast 2 deaths, wight will cast 2 curses instead of 4.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -104720,7 +104720,7 @@
     var goblin = new Goblin()
     {
         MaxHealth = 43, Health = 43,
-        MaxMana = 16, Mana = 16,
+        MaxMana = 12, Mana = 12,
         BaseDodge = 14,
         HideDetection = 22,         
         Experience = 1306,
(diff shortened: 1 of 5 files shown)


sproatba:
seems like a decent start to some of the caster damage,  when those mm casters decide to not use the other spell they can really dish out some damage.  OG the used to cast 1 mm then area effect repeatedly now they often do 2 or more mm, really ups the damage.  we could also put some delays on there to prevent it from being cast multiple times etc.

Chilly021 (approved):

zrikzlok (approved):

PR #1557, opened 2022-04-06
Add Description for Fire and Ice Amulet

Chilly021:
Currently it's blank so it does not even have a description. When identified it just says 3 of 3.


sproatba:
was looking this over and noticed the f/i amulet default constructor is passing in the art for the pink pearl locate amulet, do you want to add that to this pr as well?  they both should be 5

Chilly021:
> was looking this over and noticed the f/i amulet default constructor is passing in the art for the pink pearl locate amulet, do you want to add that to this pr as well? they both should be 5

Thanks! was just looking for description. Updated with your find.

jkachhad:
The current modification actually makes it 5 charges, does not change the art. Need to change that on base(5)?

Chilly021:
image

sproatba:
good call I missed the this vs base interaction there.

yarrim (approved):

SewerCzar (approved):

PR #1564, opened 2022-04-07
add banker jeweler ring to UK Vendor

Chilly021:


SewerCzar (commented):
I've been thinking since there are major economic impacts to these items - I would make them cost a fair bit of gold each. Like 1-2 mill a piece. I feel like a bit of a grind for items that impact a characters economy is pretty fair.

Chilly021:
While I understand your sentiment, I am just moving it and bumped it to 5k. Only difference I see to this is people don't have to recall again and again making it a quality of life improvement overall. If this was something to decrease the difficulty of an entire segment or make something trivial, I would agree with you. I think this needs a community discussion hence the PR to get the item out there. My hope is we can make UK itself a restricted portal. (17+) The only issue is as I mentioned these are not bound items. So anyone can go out and buy for anyone. A community bank could occur making 1 million gold completely trivial as well. Lots of things where I am leaning on a healthy discussion similar to item progression, exp progression, or current economy revamp discussions.

Chilly021:
image
The poll closed today with 10k leading the front.

Chilly021:
Both have been made up to a million by community vote. @SewerCzar @jkachhad @Leolus @yarrim @dedmist @zrikzlok @Tawnos-LOK. 

Seems pretty fair here. Let's maybe think about quest ideas for the next items. Moving off of maybe doing Escorts and maybe kill quests similar to CGs but more fun and less static code.

SewerCzar (approved):

yarrim (approved):
@Chilly021 As written, this change will have the jeweler stock each of these rings 1/7th of the time per facet, since the actual items stocked will be 3 of the 7 potentials per reset.
IF the jeweler stocks one of these rings, there will be 100 available, more if allowed to restock.
I'm not mad at that, keeps people potentially looking for them and not just guaranteed availability, but that may not have been your intended goal.

Chilly021:
Oh.. we don't need that. I didn't switch the percentage.

PR #1565, opened 2022-04-07
Calculate a figure for XP per Threat and Health

TimPickup:
The idea is to give a figure we can compare and see at a glance if something may currently be giving out too much or to little XP based on its stats.

The resolution of an anomaly could be to change a combination of the XP, or the Skill or the Health of  the creature to balance it.

Initially a value around 3.5 seems to be about in keeping. Values off this May want looking at.. but this is just a tool and its far from exact. It can sometimes highlight a creature has Something wrong.

Currently Magic threat values are approximated on "Magic" values and these could need adjusting.

There are also a couple of improvements that could be made in the calculations if this is proving useful and wanted.

As this doesnt already exist as such - it should not adversely effect the editor, its simply new information.

### DIFF
--- a/Source/Kesmai.WorldForge/Editor/MVP/Models/Game/Entity.cs
+++ b/Source/Kesmai.WorldForge/Editor/MVP/Models/Game/Entity.cs
@@ -103,8 +103,35 @@ namespace Kesmai.WorldForge
 			}
 		}
 
-		[Description("Approximate offensive power (melee,ranged & magic)")]
-		public Tuple<int?,int?> Threat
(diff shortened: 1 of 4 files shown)


SewerCzar (approved):

Wirednuke:
Made something similar in excel and I was exporting the segment files to it... this is way cleaner. I like it. Wish I could approve.

Wirednuke (approved):

Chilly021 (approved):

yarrim (approved):

PR #1567, opened 2022-04-09
Scaling back some resists on Thaum pets. We are still testing. Need to find an even ground. 

Chilly021:
Then hopefully we can get some scaling and xp from them!

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Phantasm/SummonedDjinn.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Phantasm/SummonedDjinn.cs
@@ -18,7 +18,7 @@ namespace Kesmai.Server.Game
 			BaseDodge = 25;
 			Mana = MaxMana = 24;
 			IceProtection = 90;
-			MagicProtection = 50;
+			MagicProtection = 25;
 
 			Movement = 3;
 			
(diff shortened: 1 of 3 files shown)


sproatba:
not sure if we have access currently, but something along the lines of hp = 25 * skillLevel  or whatever number works out right, and possibly do something similar with the resist would be a good fix here.  Right now, we are kind of stuck picking a single number to cover the entire existence.  Setting phantoms weaker while setting dj and efreet stronger helps alleviate that some, but also means the thaum can't pick the best tool for the job, but rather needs to pick dj/efreet and never use phantasm once the others are unlocked.   I think we can also dynamically set resistances based on the spell passed in so ice dj could be immune to ice while conc dj could be immune to conc but not require them both to be.

Chilly021:
The issue with what you are proposing is there are 3 classes to handle all of it. 

1) Summon Phantasm
2) Summon Djinn (Ice / Conc)
3) Summon Efreet (DB Fire / DB WW)

So if get more exposed there, maybe? I think overall Lux will just need to dynamically tune them. This was only to get Thaums to a place so they are not so damaged playing them. Right now they are over tuned, this helps the pets, but not the thaums.

yarrim (approved):
I see no problems with tweaking these values. Lets ensure any Thaum players are given the warning that their summons get weakened.

Chilly021:
I think for now we will leave these until we have Lux put these in. I had a few conversations about this. Although strong for early game, they are "okay" for late game.

Issue #1079, opened 2021-11-17
Broken corner walls AI


stormhalter:
On hexes where two walls meet, a player can step on the hex if one wall is broken but mobs still treat it like a wall is active.  This is kind of wonky and completely exploitable as if it is a pre-nerf illusion hex. [Reported by Vernak]

jkachhad:
How can you have a wall where only 1 part of it is broken? Unless it's build that way into the editor. Player should also not be able to step on it.

Tawnos-LOK:
Yeah, where there is a horizontal wall meeting a vertical wall on a hex, concussion can break one and not the other.  This allows for player movement onto the hex but not mob movement.  Additionally, a player can set a recall on a wall hex one day (via portal or breaking) and recall back onto the intact wall the next day after server reset.

PR #1568, opened 2022-04-09
exploit square - nothing came here

TimPickup:
This square near lock pick town, when turned to rubble, you could stand on it and no creatures would would you there.

There was 1 part destructible and 1 part indestructible which may have confused it so I have made it so both parts are indestructible.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -33464,6 +33464,7 @@
           <wall>158</wall>
           <destroyed>161</destroyed>
           <ruins>170</ruins>
+          <indestructible>true</indestructible>
         </component>
         <component type="WallComponent">
           <wall>157</wall>


TimPickup:
Was:
image

Now:
image

Chilly021 (approved):

SewerCzar (approved):

sproatba:
good catch, though I had taken care of all these, but if we run into anymore they either both need to explode or neither.  happens alot along the southern and eastern walls especially.

TimPickup:
> good catch, though I had taken care of all these, but if we run into anymore they either both need to explode or neither. happens alot along the southern and eastern walls especially.

when you say taken care of do you have a separate PR? as its not fixed as of right now in game. If so should I close this PR @sproatba

Chilly021:
@TimPickup  I think he meant he had already created PRs to address the majority of the issues. He missed this one, and he thanks you for finding it. This PR still is good.

sproatba:
yeah, bad typo though = thought  I missed this when I went through and cleaned everything up.

yarrim (approved):
Nothing wrong with this change, it keeps the walls consistent, but the 'exploitiness' was addresses in issue #1079

PR #1569, opened 2022-04-09
Resolving Art issues.

Tenscore:


Chilly021 (approved):

SewerCzar (approved):

yarrim (approved):

yarrim (approved):
I thought the source element might need to be adjusted, but after reviewing the #artwork channel, the image size stayed the same, only the art now fits in a 45x45 region anchored to the bottom right.
Carry on.

PR #1574, opened 2022-04-11
Move only banker ring to UK and make it so that there are 7 every restock

Chilly021:


Chilly021:
Quoting @yarrim  "As written, this change will have the jeweler stock each of these rings 1/7th of the time per facet, since the actual items stocked will be 3 of the 7 potentials per reset.
IF the jeweler stocks one of these rings, there will be 100 available, more if allowed to restock.
I'm not mad at that, keeps people potentially looking for them and not just guaranteed availability, but that may not have been your intended goal."

sproatba:
yeah, so in this case there are 6 items so he'll have it roughly half the time and 7 of them.  the default on restock is true so unless I'm misunderstanding this, he'll restock the rings throughout the day when he does have them.   Just making sure you are getting what you are looking for.

Leolus (approved):

SewerCzar (approved):

yarrim (approved):

PR #1584, opened 2022-04-15
Wizard Magic Hat

Wirednuke:
image

Created a Wizard Hat that provides:

**10 Stun protection** - Figure we can debate this factor
**Night Vision** ,- No one will use it if it doesn't have this.
**15 Magic Protection** - The griffin skull already gave 15 to fire and ice, this will just make it 15 across the board. 

Also... it's Wizard only. Thought about making it work for thaums too.. but I figured we could discuss that. Wizards need it since they don't have big boys in blue to hide behind. 

As for drop locations.... OL.. Vlad.. Bloodlands somewhere? 

This is a collaboration with Tenser, he's taking care of the art side, but see example above.

### DIFF
--- a/Content/Localization/Enu.xml
+++ b/Content/Localization/Enu.xml
@@ -284,6 +284,7 @@
   <entry index="6000101">feather</entry>
   <entry index="6000102">idol</entry>
   <entry index="6005000">corpse</entry>
+  <entry index="6000103">Hat</entry>
 
   <entry index="6100000">{0}</entry>
   <entry index="6100009">You can't do that here.</entry>
@@ -691,6 +692,7 @@
   <entry index="6200351">The gem turns to gold in your backpack.</entry>
   <entry index="6200352">The gem turns to gold at your feet.</entry>
   <entry index="6200353">a large gnarled staff with a ram skull attached.</entry>
+  <entry index="6200354">a spectacular azure wizards hat with bright yellow stars.</entry>
 
   <entry index="6250001">The armor seems quite ordinary.</entry>
   <entry index="6250002">The combat adds for this weapon are +5.</entry>
@@ -825,6 +827,7 @@
   <entry index="6250131">Inside you see a red liquid.</entry>
   <entry index="6250132">Inside you see a clear liquid.</entry>
   <entry index="6250133">Inside you see a yellow liquid.</entry>
+  <entry index="6250134">The hat emenates power, and shimmers as you touch it. It contains the spell of Night Vision.</entry>
  
   <entry index="6300001">You feel a momentary vibration in the {0} as you lift it.</entry>
 
--- a/Source/Kesmai.Server/Game/Items/Equipment/Helmets/WizardHat.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Helmets/WizardHat.cs
@@ -1,0 +1,54 @@
+using System.Collections.Generic;
+using System.IO;
+using Kesmai.Server.Network;
+
+namespace Kesmai.Server.Items
+{
+	public partial class WizardHat : Helmet, ITreasure
+	{
+		/// <inheritdoc />
+		public override int LabelNumber => 6000103;
+
+		/// <inheritdoc />
+		public override uint BasePrice => 3000;
+
+		/// <inheritdoc />
+		public override int Weight => 80;
+		
+		/// <inheritdoc />
+		public override int ProtectionFromStun => 10;
+		
+		/// <inheritdoc />
+		
+	    public override int MagicProtection => 15;
+		
+		/// <inheritdoc />
+		public override bool ProvidesNightVision => true;
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="WizardHat"/> class.
+		/// </summary>
+		public WizardHat() : base(35)
+		{
+		}
+		
+		/// <inheritdoc />
+		public override bool CanEquip(MobileEntity entity)
+		{
+			if (entity is PlayerEntity { Profession: Wizard })
+				return true;
+
+			return false;
+		}
+		
+
+		/// <inheritdoc />
+		public override void GetDescription(List<LocalizationEntry> entries)
+		{
+			entries.Add(new LocalizationEntry(6200000, 6200354)); /* [You are looking at] [a spectacular azure wizards hat with bright yellow stars.] */
+
+			if (Identified)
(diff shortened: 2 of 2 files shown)


SewerCzar (commented):
As things stand I firmly believe that adding more items is not the way to go. I believe there've been a bunch of discussions on that.

I actually like the concept of the hat I just dont think it's a good move right now

Wirednuke:
> As things stand I firmly believe that adding more items is not the way to go. I believe there've been a bunch of discussions on that.
> 
> I actually like the concept of the hat I just dont think it's a good move right now

I understand... and I see this as a band-aid while we wait for a more permanent fix to the issue. With the code not being exposed, and Lux being busy in RL, it will be a while before that happens. We have a class that isn't getting played much in the meantime, and I see this as a quick and easy fix to one of the higher priority issues. 

ALSO: it's an easy and easily revertible way to put theory to practice before Lux makes a larger sweeping change to the mechanics of the game. 

I genuinely appreciate your input though.

sproatba:
we have 2 useless skulls to figure out before adding new items imo.  I could definately see hats being a thing we could work in though.  trade off stun for something else maybe

SewerCzar:
More people are playing wizards than thieves this last week do we need that to be high priority too? 

We can't have items being added because a few people in the community decide they want a band-aid. The last time we had a community driven "class balance" it led to the severe imbalance we have with iceys right now. 

We've had conversations about class balance and the ball is in Lux's court. And honestly with the pace of HC vs official live there's no "high priority" issues balance wise.

Chilly021:
Please do not close PRs that are not yours.

SewerCzar:
> Please do not close PRs that are not yours.

There are 8 contributors. It only takes 3 approvals. There is no disapproval. All this has to do is stay open long enough and people will approve it. This doesn't need to be open until that happens.

dedmist:
I think this could help test balance, it's probably not the right long term fix, that should be magic skill based, but I'm not opposed to testing. 
I would like it if we had a process to determine closures after more discussion.

dedmist:
> > Please do not close PRs that are not yours.
> 
> There are 8 contributors. It only takes 3 approvals. There is no disapproval. All this has to do is stay open long enough and people will approve it. This doesn't need to be open until that 

It does seem like we get people auto approving to clear a queue from time to time,  but does this happen if changes are requested?  We should hold approvals if change requests are on the table, treating it as a hold for discussion at least.

Wirednuke:
What if I add this item to segment code In blackburrow so we can test it.

If it works I can utilize the same stats on the bear skull since it's not being utilized and we can save the wizard hat idea for blood lands itemization.

sproatba:
If this goes in on a non HC land for testing I'm game.   I think we could also try it as the dragon skull upgrade.

yarrim:
> > > Please do not close PRs that are not yours.
> > 
> > 
> > There are 8 contributors. It only takes 3 approvals. There is no disapproval. All this has to do is stay open long enough and people will approve it. This doesn't need to be open until that
> 
> It does seem like we get people auto approving to clear a queue from time to time, but does this happen if changes are requested? We should hold approvals if change requests are on the table, treating it as a hold for discussion at least.

The way *I* approve is by:
* Code correctness: is this going to compile, Or at least do I think it will. If this is segment code and I am providing approval #3, then I !compile to ensure it won't break at restart.
* Is this appropriate: does the submitter have a reasonable argument for the change. I'm not really keyed into balance, so I assume that our contributors are working in the best interest of the game/community. I hopefully will notice wild balance issues, but even if I don't, we can adjust going forwards.
* Is this needed: I've stayed out of the wizard hat PR until now because I'm not sure it's necessary. Angry and Brick are swaying me that it could be added as a SC item somewhere just to see if it works, but I also agree with Sewer that we have to otherwise useless helms that could use some attention. I also like the idea of different gear that is somewhat class specific that serve similar purposes: wizard hats for F/I & NV and knight hats for F/I & NV, so I'm personally torn - and thus have let others make their thoughts known.
* Is there a discussion: Once there's even just one person advocating to reject or to change or adjust it in some way, I feel the change probably needs a broader discussion before getting blanket 'correctness' approval. In Github, this is easy to see by the comment indicators in the list view. If I have thoughts, then I'll add them. If I see something like a consensus, then approval is possible.

All contributors have the ability, I think, to edit PRs. Let's get in the habit of marking them with either labels or some text tag like [discussion] to indicate changes we think need more attention than just yes/no votes.

Chilly021:
> > Please do not close PRs that are not yours.
> 
> There are 8 contributors. It only takes 3 approvals. There is no disapproval. All this has to do is stay open long enough and people will approve it. This doesn't need to be open until that happens.

There is a way to request changes with comments though. Flat out rejecting something out of fear of approval is a faulty way to look at things. It stifles a hearty commentary. 

Now to issue at hand. We have class imbalance issues. If another person blames it on HC, I think they are only seeing half the actual issue then. Hardcore has exposed a lot of things that normal play wouldn't have really been able to do. 

We have played Kesmai through Underkingdom hundreds of times over now. We are seeing certain classes fly to 20. Others though are held back due to their lack of HP and lack of Block. 

The Magic Hat, for example, I could see something like this for Ydmos's twin or something in AG. The issue with trying to add new items at this point to BG is it will offset a lot of things potentially, even if it's made usable by one class. 

We could add them now, but I think Lux will need to come back and add the true class fixes in that are looming. We have half implemented classes in some directions. Others are now having to be pet only classes. Wizard at this point have to creep around or cast 2+ hexes back in ianta room with whirl winds just to not be seen and have it easy. Wouldn't you rather walk around and cast while fighting? Comfortable play to okay for some, but others like to press their luck and try for other areas where dark/djinns/stun/fear/blind/hide is useful in tight spots. 

If you play any of the new player created content, you know casters are reduced to battle mages and level up the one weapon that hits the hardest and the other that blocks the best. I could see a ton of people using BBS late game instead of an actual staff due to out leveling the need for +mana on a OL staff. Plus OL staff here hits harder than it did in any other version. It was effectively a wet noodle in other versions but could block pretty okay. (AMAZING in OG) 

So again.. adding this hat, I see a lot of testing could be applied. Do I agree it should be added to Hard Core accessible areas? Not without being tested in SC first. So add it to SC areas where people can test to their heart content. 

Then if acceptable maybe then we add it to HC. Heck even bloodlands.

Not going to have another dagger vs dagger stand off of item additions here though. Otherwise I'm putting in Wings where I can fly over air hexes and fight on them.

Wirednuke:
Closing this for now, will include it in segment code in blackburrow for testing at some point.

PR #1585, opened 2022-04-16
Hummingbird sword quest

TimPickup:
This is a slight mix between what originally happened and what happened in lands. I can't be sure all of these mechanics will work as expected as they haven't been done like this before so it will need testing and possibly revising but I can't test it without putting it live.. and currently hummers are just free with conc bottles so this should be a good improvement. 

IMO this is a battle someone that has duoed TK should be able to solo (as its only soloable) so shouldnt be "that" hard. Just a check you are not a low level.

I believe I have managed to make it so only one person can enter the room at once. You then have to enter the centre of the room and defeat the guardian who throws WWs and hits you to check your fortitude.

This is what I changed:

1) removed normal portal hex
2) replaced with a special invisible entity that on entering their square, checks if you are holding a hummer amulet And if the guardian is available and alone (no other players)
2.1) If the guardian is not spawned currently then the player is refused entry
2.2) If someone else is already in the guardian room with the guardian then the 2nd player is refused entry. As per the original game this is a solo challenge to show your basic fortitude
2.3) If the above pass then you art teleported in and the amulet is consumed
3) the guardian is only attackable by physical means. No magic and no conc bottles etc. This is meant to be a basic show of fortitude.
4) The guardian will only start attacking when you enter the centre of the room. From that point it will be able to move and cast etc.
5) Guardian has a weak WW but its instant cast and will cast 2 in quick succession for effect. The combined effect will be a reasonable tick of damage but wont last too long. The attacks then are again reasonable but not too bad.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -104251,6 +104251,9 @@ var gargoyle = new Gargoyle()
       <entry entity="uktownConfessorGhost" size="1" minimum="1" maximum="1" />
       <location x="15" y="3" region="2" />
     </spawn>
+    <spawn type="LocationSpawner" name="Humming Portal">
+      <minimumDelay>0</minimumDelay>
+      <maximumDelay>0</maximumDelay>
       <script name="OnAfterSpawn" enabled="false">
         <block><![CDATA[]]></block>
         <block><![CDATA[
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -205,7 +205,14 @@
         {
             Name = "guardian";
             Alignment = Alignment.Chaotic;
-            Body = 45;
+            Body = 26; //air elemental body
+        }
+        
+        /* Called when someone */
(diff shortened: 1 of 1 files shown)


Chilly021 (commented):
I'll have to check this out either late tonight, or tomorrow. Loving the idea, just making sure to give a look on the world forge. It'll require me to pull it down.

SewerCzar (approved):
Checked the code and I believe all of it should work. Has some good implications for future encounters if it does.

I love it.

dedmist (approved):
I'm approving because I assume there needs to be some balance changes, and there will be a second pass,  but I'd like to see a few changes:

Kes messages should be a little cryptic IE "This area flutters with an energy you can't comprehend"

What makes the fight more interesting is that ww is cast at random, forcing you to hide behind tables and jump around, whereas the ww here would be predictable based on time.  It would be nice to add ww as a random attack instead of mp based.

Leolus (approved):

PR #1587, opened 2022-04-16
Enhancements/oakvael serp level

Chilly021:
1) Established Serp Level had way too many casters and salamanders
2) Reduced casters to have 2-3 per quadrant (NE/NW/SE/SW)
3) Upped the Orc/Hob pack ratio to account for density readjustment
4) Fixed an issue that could arise if Tim.the.enchanter AoEs the other lair crits and they turn around/kill him. OnDeath had a "

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -102289,7 +102289,7 @@
         <block><![CDATA[]]></block>
       </script>
       <entry entity="level9Hobgoblin" size="3" minimum="12" maximum="15" />
-      <entry entity="level9Orc" size="3" minimum="14" maximum="16" />
+      <entry entity="level9Orc" size="3" minimum="12" maximum="15" />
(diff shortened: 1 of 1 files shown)


SewerCzar (commented):
There's an extremely low density left over from the fact that Lux used lands values and then the level was expanded 3x. Need a massive increase in density on Serp level, if you look at the spawns it's extraordinarily low

Chilly021:
Welcome to add more to it. I just know hunting there turns into caster bundling throw us some numbers and we can add it to this PR.

Chilly021:
I'll show a screenshot where I had 6+ casters on screen due to crit bundling. It generally happens when a player just enters a segment and someone might have been hunting there before. I had 3x sallies and 6+ casters all going to war with one another.

SewerCzar (approved):
Fair enough this pulls fine just will need another in the future to make serp more attractive to hunting

sproatba (approved):

Leolus (approved):

PR #1594, opened 2022-04-17
Added immunity to base vendor model

SewerCzar:
Will go back and clean up NPCs later to remove individual immunities

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -107270,6 +107270,7 @@
     {
         Name = _names.Random(),
         Body = (female ? 10 : 9), IsFemale = female,
+        IsInvulnerable = true,
         
         BaseDodge = 10,
         MaxHealth = 50, Health = 50,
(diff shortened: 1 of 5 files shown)


Chilly021:
Not sure the purpose around this? We removed some of the immunities originally due to scripting in peaks. Vendors with immunity just sounds like we will have a few people just casting without adventurers bonus. I'd prefer them have like 1 health and die personally.

SewerCzar:
> Not sure the purpose around this? We removed some of the immunities originally due to scripting in peaks. Vendors with immunity just sounds like we will have a few people just casting without adventurers bonus. I'd prefer them have like 1 health and die personally.

Every vendors stock is reset when they die.

 It's just notably that UK ones are being manually restocked right now. And we can't have "rare goods" when players can force reset,

Also peak invulnerabilities were never removed, and I added the invulnerability to axe/kes ones post-peaks

Leolus:
I think vendors should be immune and just not grant skill on attack. Either way most vendors are already immune so this change doesn't really impact the current state of the game, just makes it easier later on so you don't have to remember to do it for each vendor.

Leolus (approved):

Chilly021:
I think you disabled the Thaums up front in town due to scripters were using them as meat shields.

yarrim (approved):

PR #1595, opened 2022-04-17
Wendy Update

SewerCzar:
I actually like Wendy's whirlwind damage where it's at - it can be healed through.

Feedback is that there isn't enough threat on her with the once per minute cast.

I propose reduction of total mana to let her cast 3.33 times per minute.

Threatening but easily played through

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -98839,7 +98839,7 @@
         Name = "dragon",
         
         MaxHealth = 2100, Health = 2100,
-        MaxMana = 20, Mana = 20,
+        MaxMana = 6, Mana = 6,
         BaseDodge = 28,
         
         Experience = 85594,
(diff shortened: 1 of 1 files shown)


Chilly021:
That's fine, I was sort of loving the old WW where it could kill you. Either way.. it's fine though.

Chilly021 (approved):

sproatba:
What am i missing?  6 mana takes 12 rounds to regen.  12 seconds on HC or 5 times per minute.  36 on agate or .6 times per minute.   where does 3.33 come from?

SewerCzar:
> What am i missing? 6 mana takes 12 rounds to regen. 12 seconds on HC or 5 times per minute. 36 on agate or 1.6 times per minute. where does 3.33 come from?

For some reason mana regen wasn't working in a way I particularly understand, but what I do understand is this:

Wendy had 20 mana and was casting whirlwind again on the 60 second mark.

If we divide the number by 3.33 it'll be 3.33 times more often than 60 seconds (In theory less because of the mana stoppage when casting one of the three times)

So 3.33 more times per minute

yarrim (approved):

PR #1598, opened 2022-04-18
Princess Quest Changes

SewerCzar:
In this pull I created a GoogogVendor entity that sells PrincessStones for 50k gold. 

I also created the PrincessStone based on a template that works in another land - it will be a gem that can fit into backpack. It's using the yttril model right now. 

I've changed the Jailer into a Distracted Jailer shopkeeper - It should work as other monsters don't fight shopkeepers and the onincomingplayer trigger works for them.

Then when a player walks onto the hex it searches for the PrincessStone in either hand, if it doesn't have it it returns. If it does have it it deletes the item and opens the gate and despawns the jailer. 

Every component individually works and has been tested so I'm fairly confident. It's also a big improvement to craft princess craft scroll nonsense I made before.

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -616,6 +616,65 @@
             }
         }
     }
+    
+    public class PrincessStone : Gem
+    {
+        /// <inheritdoc />
+        public override int LabelNumber => 6000075;
+
+        /// <inheritdoc />
+        public override uint BasePrice => 5000;
+
+        /// <inheritdoc />
+        public override int Weight => 500;
+
+        /// <inheritdoc />
+        public override int Category => 3;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrincessStone"/> class.
+        /// </summary>
+        public PrincessStone() : base(110, 5000)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrincessStone"/> class.
+        /// </summary>
+        public PrincessStone(Serial serial) : base(serial)
+        {
+        }
+
+        /// <inheritdoc />
+        public override void GetDescription(List<LocalizationEntry> entries)
+        {
+            entries.Add(new LocalizationEntry("A rock with that sparkles with intensity. It's very shiny."));
+        }
+
+        /// <inheritdoc />
+        public override void Serialize(BinaryWriter writer)
+        {
+            base.Serialize(writer);
+
+            writer.Write((short)1); /* version */
+        }
+
+        /// <inheritdoc />
+        public override void Deserialize(BinaryReader reader)
+        {
+            base.Deserialize(reader);
+
+            var version = reader.ReadInt16();
+
+            switch (version)
+            {
+                case 1:
+                {
+                    break;
+                }
+            }
+        }
+    }
 ]]></block>
     <block><![CDATA[]]></block>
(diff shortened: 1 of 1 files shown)


Chilly021 (changes requested):
Putting this one on hold until we get the compiler back. A bit risky otherwise.

SewerCzar:
So let's look at my code slightly in depth and see if there's any issues: I have confidence it will work but hopefully a closer eye will help:

https://pastebin.com/vW6QDYh2

This is the new princess stone. I have an example here that is the same code format: 

https://pastebin.com/BkgxvRRu

A beyond compare of the two files shows the only differences are name and description:

image

For the TrollGuard code the entity is a normal character, shopkeeper template so other monsters will not attack him:

image


        RangePerception = 0,
        
this line makes it so he only sees players on his hex. This is important for OnIncomingPlayer, it will only trigger when a player enters his hex.

Now the code for OnIncomingPlayer is a little harder to read, but I have working code to compare it to:

image

The first line of code highlighted is visible under here. The beyondcompare only flags amethyst vs princesstone as seen below:

image

The tile transformation code can be seen in the current troll jailer segment file:

image

Now for the last bit is Googog selling stones as a balm seller would. Here's a comparison of an edited balm vendor in Raz:

image

ClearBalm defines the item they sell, 600 defines the price (it's customized from 16, this is the Lab balmer). 10 defines how many per initial coin drop they'll make (we can see on normal balms if you drop 300 gold it spawns 160 gold of balms instantly).

I hope this explains everything in my pull. The code all exists elsewhere. The only problem would be if I didn't dot my i's and semicolon my t's properly.

Chilly021 (approved):

Chilly021:
Sorry was in the middle of talking to the Mrs. last night. I was fine with this PR, I just was waiting on the stability to be back. My issue is some of these things are harder to test without a test server. So although syntax wise it looks fine and programmatically it should get what you need, I don't know how to User Acceptance test this until it has already been released.

SewerCzar:
> Sorry was in the middle of talking to the Mrs. last night. I was fine with this PR, I just was waiting on the stability to be back. My issue is some of these things are harder to test without a test server. So although syntax wise it looks fine and programmatically it should get what you need, I don't know how to User Acceptance test this until it has already been released.

I mean now it's you drop 50k on googog, walk onto the troll jailers quest with the stone in hand and the princess is open.

Right now it's asking googog to craft scroll, then walking to the jailer and asking him to craft princess.

We can make it a key and an invisible NPC so when you walk on the hex you "unlock" the door automatically. With the new definitions we can do that if it's preferred.

SewerCzar:
> I've run through this quickly and so far it looks pretty good, there's a couple things I want to pull up on world forge to check because the don't look quite right to me in the github interface. I am questioning creating some of these things in segment code if segment code is so buggy to manipulate. Should we be putting these in the cs files, waiting for a lux compile then adding them to segment?

Segment code isn't necessarily buggy it's just items created via segment can be. Just clarifying as your point is taken.

Once we get access to the item entity file with categories we can make a proper key in there and when we edit it it, in theory, won't destroy the server.

I'll put up a pull once we have the item file to category 3 key, where googog sells it and when you step on the hex an invisible entity opens the door. I think everyone will agree that's a good way to do it.

This pull was mostly because people were frequently lamenting the scroll being stuck in hands.

PR #1599, opened 2022-04-19
Add indestructible door in ianta

Chilly021:


SewerCzar (commented):
Not really sure what this solves. People will still free hunt up there as knights using the door, and wizards just have to find secret door occasionally.

Chilly021:
Since the room really isn't supposed to be a "dangerous" room due to ianta, we've seen it blown out pretty much every morning/day. It has been hunted by quite a few of us and most have came to the same conclusion, just make the door indestructible if it's for ianta in the future. 

I can start designing the quest, but the it will have to have some different updates. For example locked doors are not a thing. and ianta in lands offered a key on rescue to the door next to power troll. I think OG was a random stat pot.

Maybe this time we can think about we can think about a DP or something random for the drop.

Leolus:
I feel like a better compromise here is to add another door somewhere so Wizards can still blow the door and WW hunt while others can use the door as a throttling method. I don't want to systematically lock Wizards out of AoE hunting by making everything indestructible. I get the annoyance but we've already killed 300, speedy, and some other AoE hunting areas.

Chilly021:
I don't know what a door has anything to do with it though? You can still find secret door and then cast?

Chilly021:
Illusions, darkness, bonfire can all do the same.

Chilly021:
Don't make me play my Wizard. I'll show you How Tos on this stuff.

sproatba:
power troll had the key to his door, it was the same key used for the princess quest in uk.  Nothing to do with ianta
she was just a dp/wisdom og dp/will lands.  Not sure if there was some older lor on why he had a key to a portal that dumped you into a broken portal complex in serp level, but overall it was just a noob trap

sproatba (approved):

Chilly021:
Let's discuss this again when ianta quest is done. Maybe too soon to discuss.

PR #1600, opened 2022-04-19
The old mama couldn't be evaded

Chilly021:


SewerCzar (approved):
This doesn't actually do anything since her lair doesn't let her move more than 10 hexes but it's excess code so it should go

Chilly021:
She's been at the rope on multiple occasions this past week. Not sure what you mean here.

SewerCzar:
> She's been at the drake rope on multiple occasions this past week. Not sure what you mean here.

Right and tethering makes mobs not move more than 20 hexes away from their spawn. She's already hanging on the drake rope so the existed tether doesn't leash her in any way. Still good to remove. Just not going to change anything.

Leolus (approved):

sproatba:
is this going to cause her to keep roaming the entire area after she has been pulled?   If the goal is to have her fly up/down I believe lux will have to do some coding for that.

dedmist:
I think if we're going to do a full roam, then we should be doing a location spawn instead of a region spawn.
In OG, Mama was set to mostly roam the 3x3 dirt hex to protect her horde.  She would occasionally visit the Yeti, she would just fly up sometimes if she heard noise, but not venture in unless she saw something.  That was a change made near the end of the game though, as well as random casts instead of moves when no player was visible, which included the drake ranging to Mama as well.

For most of the game the ledge was 100% safe, this is where we would wait for a party to gather to kill Mama.

I don't think more roaming is a better way to go, especially if we make the Mama fight more interesting, we still need a party gathering spot.  The search mechanics already facilitate Mama looking for someone that's ducking around a corner for safety.

SewerCzar:
> I think if we're going to do a full roam, then we should be doing a location spawn instead of a region spawn. In OG, Mama was set to mostly roam the 3x3 dirt hex to protect her horde. She would occasionally visit the Yeti, she would just fly up sometimes if she heard noise, but not venture in unless she saw something. That was a change made near the end of the game though, as well as random casts instead of moves when no player was visible, which included the drake ranging to Mama as well.
> 
> For most of the game the ledge was 100% safe, this is where we would wait for a party to gather to kill Mama.
> 
> I don't think more roaming is a better way to go, especially if we make the Mama fight more interesting, we still need a party gathering spot. The search mechanics already facilitate Mama looking for someone that's ducking around a corner for safety.

This doesn't change her boundaries at all. She won't leave her spawn zone, which you set, until she sees people. Then when her pathing takes her into the boundary of her spawn region again she'll stay there. It's very simple AI we have here and again, this changes nothing. All tethered does is mean 20 hexes away from spawn they will stop / return home. That is all.

Also in this hypothetical where we set her to fly up to yeti she would remain stuck there because of sleeping segments / player entity requirements to wake it up.

Chilly021:
I believe Mama used to fly up when Stormhalter first started. Lux had to apply leashes due to player complaints due to grieving with Drake and a few others. Now Drake roams free. I'm not saying this tether will do that, but will make sure she can move freely once out. 

The AI changes a bit here at least. It might not much.

PR #1615, opened 2022-04-20
Attemp to fix princess scroll so it can be sacked

sproatba:
I know there is some work to overhaul this, just wanted to see if this works for experimentation purposes

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Underkingdom" version="0.94.0.0">
+<segment name="Underkingdom" version="0.95.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

SewerCzar:
> I know there is some work to overhaul this, just wanted to see if this works for experimentation purposes

Changing items in segment is the cause of a lot of the crash. Half the time I've edited an existing item it's created a server crashing bug that causes server to come up NPCless. I'm comfortable doing this if it's added on a night Lux is around. Can we confirm he'll be on at 1 am?

sproatba:
Works for me, if this is dangerous let's wait for Lux.

Chilly021:
I'm only approving this because... It would be nice to have scroll in bag. Though if we have an alternate fix .. just let us know

SewerCzar:
> I'm only approving this because... It would be nice to have scroll in bag. Though if we have an alternate fix .. just let us know

The problem is that segment item change code is prone to bugs. Let me write up why my code works let's see if you guys have confidence (it fixes the whole craft, scroll craft, princess thing as well)

SewerCzar:
> I know there is some work to overhaul this, just wanted to see if this works for experimentation purposes

I explained the overhaul in the other thread. I hope it makes sense. No worries if it doesnt and we can get this in place sooner rather than later.

sproatba:
closing this as we have an overhaul in progress that will be a better fix

PR #1618, opened 2022-04-22
[just bug fix] updating some spawn Max on Vlad

TimPickup:
Nothing Major. Bug fix for possible vlad not spawn + reasonable spawn time for vlad tower. Currently in HC each layer respawns in 5 mins!

Tower 1: Was 15-20 mins Now 2h30-3h
Tower 2: Was 15-20 mins Now 2h30-3h. Also changed Max spawns to 6 just because there are only 6 that are setup so 8 is impossible.
Tower 3: Was 15-20 mins Now 2h30-3h
Vlad: Was 3h-5h Now 3h-4h. Also changed max to 5 as otherwise there is a 1 in 4 chance Vlad would not spawn at all! 5 entities can spawn.

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -80274,8 +80274,8 @@
     </spawn>
     <spawn type="RegionSpawner" name="Vlad">
       <minimumDelay>10800</minimumDelay>
-      <maximumDelay>18000</maximumDelay>
-      <maximum>4</maximum>
+      <maximumDelay>14400</maximumDelay>
+      <maximum>5</maximum>
       <script name="OnAfterSpawn" enabled="true">
         <block><![CDATA[]]></block>
         <block><![CDATA[
@@ -80301,8 +80301,8 @@
       </bounds>
     </spawn>
     <spawn type="RegionSpawner" name="vladtower1">
-      <minimumDelay>900</minimumDelay>
-      <maximumDelay>1200</maximumDelay>
+      <minimumDelay>9000</minimumDelay>
+      <maximumDelay>10800</maximumDelay>
       <maximum>5</maximum>
       <script name="OnAfterSpawn" enabled="false">
         <block><![CDATA[]]></block>
@@ -80329,9 +80329,9 @@
       </bounds>
     </spawn>
     <spawn type="RegionSpawner" name="vladTower2">
-      <minimumDelay>900</minimumDelay>
-      <maximumDelay>1200</maximumDelay>
-      <maximum>8</maximum>
+      <minimumDelay>9000</minimumDelay>
+      <maximumDelay>10800</maximumDelay>
+      <maximum>6</maximum>
       <script name="OnAfterSpawn" enabled="false">
         <block><![CDATA[]]></block>
         <block><![CDATA[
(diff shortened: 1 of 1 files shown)


sproatba:
I'm totally against these time changes, part of the fight with vlad is getting up the tower and dealing with the tower if you have to jump or dealing with the top level if it respawns making the stairs unsafe.   if they need tuned fine, but the tower shouldn't be clear for the entire fight, unless you kill him quickly.  I'm also not sure why all these spawns are set for hours,  we don't have tons of facets, and they historically were much shorter.  Probably need a discussion on lair times, but 1hour spawns are a long time when you are sharing a facet, or organizing a group to farm a lair

sproatba (changes requested):
At most tower should be bumped a few minutes, it should be back before vlad is dead unless you have a group nuking him quickly.

Chilly021:
I think we are just experiencing the "Opal" speed issue right now. I don't know about changing time right now. Yes I died, and yes I knew the risk about jumping off the top and running back up. I'm not mad about it. I just was not thinking.

SewerCzar:
> I'm totally against these time changes, part of the fight with vlad is getting up the tower and dealing with the tower if you have to jump or dealing with the top level if it respawns making the stairs unsafe. if they need tuned fine, but the tower shouldn't be clear for the entire fight, unless you kill him quickly. I'm also not sure why all these spawns are set for hours, we don't have tons of facets, and they historically were much shorter. Probably need a discussion on lair times, but 1hour spawns are a long time when you are sharing a facet, or organizing a group to farm a lair

I agree with everything Angry said, including lair timers needing a bigger discussion.

The tower being murderous is the big part of the danger of the Vlad fight. If you can't kill him fast enough that the penultimate level has mobs back you might have to jump down and re-do the tower over again which is dangerous.

I think it's just part of the encounter and I personally like the concept of it.

TimPickup:
I have suggested shorter respawn times..
tower levels now 50-60 mins each .. so 16m to 20mins on HC

would that be a better reasonable respawn time ?

Chilly021:
Be good time come up with a better time. I think 10 minutes is better than 5 on opal though.

sproatba:
I'm thinking leave it as is right now, he just got a new toy that people want so he'll be seeing some traffic.  We can get some feedback to make a better decision with.

TimPickup:
ok. May as well at lease put through the fixes though for how many spawn. 

Vlad could not spawn otherwise.

vladTower2 had a max of 8 but only max 6 creatures defined so may as well also set that to 6

sproatba (approved):

sproatba:
I went and solod vlad on 18/13 knight in SC.  He honestly wasn't that bad.  Tower respawn played a role in the fight, but with the regen rate on lairs where it is that seemed fine.  If the instant cast bug gets worked out he'll be a little more difficult I think.  Overall we need to decide where he fits in the progression of the BG and tune for that.  Right now he's one of the end game lairs to be done as you are finishing the bg, but his items are a little lacking for that.

Chilly021 (approved):

SewerCzar (approved):

PR #1620, opened 2022-04-22
Normalize Axe Surface Gems with -3 gems. -3 is easier than axe surface

Chilly021:
* Upped the gem value to be "Slightly" higher than -3 gems 
-3 kesmai is based on Level 5 crits 
Axe Surface is based on Level 6 crits

image

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -104339,7 +104339,7 @@
         
     wraith.AddGold(70);
     wraith.AddLoot(new LootPack(
-        new LootPackEntry(true, AxeSurfaceGems,       4.8,     gemsPriceMutatorNormal), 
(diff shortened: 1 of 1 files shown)


sproatba:
are you also taking into account quantities here?  axe surface seems to drop a lot more gems than kes -3 so the overall gem value might be "ok".   Surface is a bit hard to judge though because so many people saturate it.

Chilly021:
I just changed the drop %. I forgot about that. Should be less gems dropping for more value.

SewerCzar (changes requested):
My personal opinion is double gem values (20k nugget 1700 emerald 3200 uncut diamond) and put the drop rate at 7.4 and I'd be happy with it, if that seems fair.

Chilly021:
Checking it now.

SewerCzar (approved):

sproatba (approved):

zrikzlok (approved):

PR #1623, opened 2022-04-23
Troll king guards quick update

SewerCzar:
TK guards had a gap in the spawn causing them to pile up in his room if people were only doing princess. This fixes gap.

image

### DIFF
--- a/Segments/Underkingdom.mapproj
+++ b/Segments/Underkingdom.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Underkingdom" version="0.94.0.0">
+<segment name="Underkingdom" version="0.95.0.0">
   <script name="Internal" enabled="true">
     <block><![CDATA[]]></block>
     <block><![CDATA[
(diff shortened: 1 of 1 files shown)


sproatba:
So I had a different fix in mind for this, but was going to do some more testing before bringing it up.  I was thinking add a spawn group specifically for the tk room and making the outer guards roam the entire complex aside from the 2 southern secret rooms.  There also seems to be some sort of issue where they won't open the 3 doors on the north side,  i've gone there 3 times now and all 3 doors were closed and was able to pull from 300 in peace.

sproatba (approved):

sproatba:
Ok that was really unclear, I was thinking one giant spawn region instead of all the small ones, then add exclusions for the secret rooms.

SewerCzar:
> 

Mobs won't cross a line missing in their spawn. I'm totally fine with one big spawn I think they got split up in the anti plink pull.

sproatba:
As I said need to test more, but it seems like if none spawn in that northern box, nothing will walk into that box even though the spawner below it is touching.  Might be an interaction with the doors or something.  This seems like a decent start though if they will walk out of tks lair.

sproatba (approved):

Chilly021 (approved):

Leolus (approved):

PR #1624, opened 2022-04-23
Small fix to quest entities alignment

SewerCzar:
Keeper was unintendedly marked as chaotic instead of neutral. Quick fix.

Seems I forgot to open this one yesterday


sproatba:
I see a dryad also got their loot nerfed, is that intentional?

SewerCzar:
> I see a dryad also got their loot nerfed, is that intentional?

Yeah. She used to be an alignment shifting boss but alignment shifts do bizarre things to AI that will have to be looked into later - it was causing crashes at one point.

Now she's just a 200 hp protector of the grove thing in case someone drags a small batch of mobs up there.

sproatba (approved):

Leolus (approved):

Chilly021 (approved):

PR #1634, opened 2022-04-26
Adding a new Test Region

TimPickup:
Adding a new Test Region currently with nothing in it, with the intention of trying new features in a more controlled environment before putting them live. It won't stop compile errors but if someone wants to try some new code and check it works before.. say, breaking the hummer portal, they could test it here first.

Each test could be put through as a separate commit so easy to just roll back to this empty state.

Initially (if this is approved) I would like to add a test in there for implementing the hummer portal correctly and also try some code to check there is only one person in an area at once. I will add that code on a separate request after this one.

You portal in from here in Kes:
image

The empty test area ready to have new stuff to test:
image

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -4505,6 +4505,13 @@
         <component type="StaticComponent">
           <static>242</static>
         </component>
+        <component type="HiddenTeleporterComponent">
+          <destinationRegion>5</destinationRegion>
+          <lightphases select="all" />
+          <moonphases select="all" />
+          <professions select="all" />
+          <alignments select="all" />
+        </component>
       </tile>
       <tile x="8" y="5">
         <component type="FloorComponent">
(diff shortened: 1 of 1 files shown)


sproatba:
should we put an item restriction on it so people don't randomly walk in there?

TimPickup:
@sproatba I did think something like that and someone else I suggested the whole idea to said they thought anyone should be able to enter to help test the content.

Its deliberately out of the way. The idea would be to put nothing in there beneficial.. just stuff for testing. So if you want to test dropping a hummer based on X Y Z then.. actually drop a longsword or whatever.. nothing useful, just testing the mechanic you want to test.

It shouldn't matter if people go there.

Equally if you want people to had to hold a Balm in their hand or something to enter then I could change that. I dont really mind

sproatba:
yeah, i think something easy to get that you won't normally have in your hand would be good.  Maybe football for some fun?
does that even tan here?

sproatba:
anyone else have thoughts on this?  crickets in here

yarrim:
I could see a brand-new player exploring west of town and stumbling into this, but even in that case, it doesn't seem like a big deal.

If we want to make the entry somewhat arcane, require holding a balm or mugwort or something trivial to acquire, but I don't think it's necessary.

sproatba (approved):

SewerCzar (approved):

PR #1644, opened 2022-05-01
Poison Protect Amulet. Add one to ianta quest to ensure others try the Leng…

Chilly021:
… Drake without having to go directly to OL. Currently a group of casters and a single Tank comes to OL. This results to 47% chance that anyone would get it and also that a caster could also get it versus the melee. Melee and casters could get it from ianta which allows the boss for Leng Drake to be unlocked. It will make sense since Water fall area => ianta and the level up from it is Leng Drake

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -374,7 +374,7 @@
             
             if (queen != null)
             {
-                new PermanentWillpowerPotion().Move(Location, true, Segment);
+                new SilverDaggerAmulet().Move(Location, true, Segment);
                 Say($"Hi Mom! What's for dinner?!");
                 queen.Elated();
                 


Tenscore:
Make sprigs and scorp amulets great again!
Seriously.  Its like they want this because other remediations are 'inconvenient'.

I'm on the nope train with this one.
Adding:  if we stop putting high level stuff in low level places people may not feel they need this high-end stuff yet.  We have tools.  Make new lands.

Chilly021:
I mean you can't do Leng Drake with just those two.

sproatba:
Honestly, this is a sign that ol, leng drake, and probably a few other areas have some tuning issues.   Really there are very few areas in the game where poison is even an issue right now, and most of those are areas you would likely want to start exploring before you are ready to go after OL.  Also, OL is really a caster lair since the rest can't even hit him.   We need to decide the game progression, and tune for that.   Right now you have to to either do OL for a pp amulet, or get extremely lucky on the other pp outlets.  There is a reasonable alternative in the scorp amulet, but the areas also seem to have heavy death casters as well meaning s/d amulet or s/d bracer which is also ol.   Ianta just isn't the right place for it, unless we ramp up poison in leng.  Something that can be done to keep some of the areas in check would be to drop sprigs or some other neutralizing agent in some areas.  I think we need to look at adding sprig sales in all lands that have poison you would care about although red recalls do mean running to kes for sprigs is a possibility.

In my opinion, grove is a perfect area to introduce a second pp outlet into the game.  You can hunt it reasonably well without the amulet, but having it makes it easier.  That is exactly the formula you want for those first areas that introduce a new threat.

Chilly021:
I'm okay with Grove getting it.

Chilly021:
Let's move this around. I'll submit a poll about the 100% on OL since that seems to be a lot of people's feedback. Thanks again!

PR #1645, opened 2022-05-02
Feedback is resistance is too strong.

Chilly021:
* Let's scale this back and wait on Lux to implement dynamic leveling. 

Again this was all a test to see the variances of things one could do with more resistance.


sproatba (approved):

dedmist (approved):
Can we get the cast level?  It'd be better for these to scale.

sproatba:
yeah, seems like scaling should be pretty easy, we just don't have the variables exposed to do it.

Chilly021:
That's the hope in the future. I would love to see it.

Leolus (approved):

PR #1646, opened 2022-05-02
Leng Drake - Discussion - Nerf or Add new stuff on it

Chilly021:


sproatba:
I agree this guy needs some tuning, but I'd like to do a little more testing on him before we decide where to place him.  Right now there is some tension between using him to get the lres ring which unlocks uk hunting (supposedly)  that leads to OL which leads to PP to make him easier.   If you choose ldrake first you don't get PP to take him down.  If you choose OL first your only option for Lres on the way down is the shield from axe drake.   Until you have conquered OL, this is one of the areas where you have to deal with poison and death.   That type of area fits the bill for end game BG content.   I guess the question is, should he be a stepping stone to uk, in which case he should be tuned for characters in the 16-18 range, or do we keep the trade off going where he should be more tuned to 18-20 range.  Right now he feels more like 21+ range to me.

Chilly021:
Closed this a bit pre-maturely.

sproatba:
We've had some conversations, and I think alot of people are leaning toward swaping the lightning shield and lres ring between axe drake and leng drake, then potentially adding something to leng drake like a pp amulet to fix progression/loot issues.   If we want to keep him a top end boss this seems like a good plan to me.  I believe we have 2 bugs related to him to fix as well.  His ar doesn't report and/or work correctly with having 3 different skill levels, and poison stacking seems to have a bug making stacked poison less deadly.

dedmist:
I don't agree with swapping ring and shield.  We shouldn't give immunity to everything too early for things that are inconvenient rather than necessary, this would be poor progression. 
There's also plenty of reason to kill the axe drake, the shield could go somewhere else, like a random lit caster drop on speedy.

sproatba:
shield and ring give the same immunity amount.  It comes from a lair who has a fairly nasty version of it so I don't really see an issue with swapping them.   We could also make one or both random drops in uk, but if we put lres ring somewhere other than leng drake people literally won't do him, just like they don't do vlad because what he offers can be found somewhere easier.   Leng drake is in a bad spot progression wise, we are trying to fix that.  Quite a few like having him more of an end game boss since he does require pp and s/d it does make some sense.  Keeping that though makes lres too difficult to achieve if we actually tune uk so you want it.  Right now even people who have it don't wear it

dedmist:
The shield requires a weapon power trade-off, it's less useful than the ring.  
If the ring were moved to axe and since the axe drake is commonly farmed, this would make lres too easily attainable and give lightning prot well before it's needed.

sproatba:
when do you think you need ligtning prot?

dedmist:
I think,  lightning and poison should be necessary in AG areas, but only convenient in BG.  Lres and pp should be the difference between solo and party on leng drake, and make the UK drake 50/50 for 18 knight without lres.
In UK on regular mobs, lit should be very survivable with area management.

Chilly021:
I'll say lightning doesn't do much right now. There is no point to the Leng Drake right now. Yes slim percentage for the plat dex. He's a risk too with his 170+ capable swings. I have videos of each. Almost 2 shot at lvl 20. So main thing is..  Drake is sort of pointless for it's difficulty. At least vlad has gem ring.

sproatba:
so I'm guessing he meant leng drake 50/50 at 18 not uk, but It sounds like he's on board with the nerf path which was my original thought.

Tenscore:
LRez shield belongs on Leng Drake.
Poison on Leng drake should not be so strong it requires PP amulet.
He is a prelude to more dangerous poisons in the UK. (Lurkers, etc.)

In the OG, Leng Drake poison was remediated by Neutralize amulets and sprigs.
He was nick-named "Sparky" for his proclivity to use lightning.

idk why everyone wants to make him more 'endgame', and push all these ancillary 'bosses' to be mightier than Mama. (Wendy went through this too.)

Can we agree that the BG endgame boss was "Mama", and the UK endgame boss was the "Overlord"?

Because that perspective helps un-conflate all of this back-and-forth a little bit.

Chilly021:
That's the issue. As it stands he is harder than most if not all bosses due to lack of PP. You can not take a sprig without risking being bolted / crit for 170+ in the current state. The nerf was the poison itself.

PR #1647, opened 2022-05-04
Some Framework for Bloodlands Itemization

SewerCzar:
Worked with ZZ to get some framework for Bloodlands itemization.

it's nothing fancy and needs a lot more work but this will enable us to link various items to worldforge OnDeath scripts and work on the potential crafting portion

### DIFF
--- a/Content/Localization/Enu.xml
+++ b/Content/Localization/Enu.xml
@@ -283,6 +283,7 @@
   <entry index="6000100">dull opal</entry>
   <entry index="6000101">feather</entry>
   <entry index="6000102">idol</entry>
+  <entry index="6000103">warhammer</entry>
   <entry index="6005000">corpse</entry>
 
   <entry index="6100000">{0}</entry>
@@ -691,6 +692,28 @@
   <entry index="6200351">The gem turns to gold in your backpack.</entry>
   <entry index="6200352">The gem turns to gold at your feet.</entry>
   <entry index="6200353">a large gnarled staff with a ram skull attached.</entry>
+  <entry index="6200354">a scythe with a blood groove leading all the way down to the handle.</entry>
+  <entry index="6200355">a sharp dagger with a blood groove that extends all the way to the handle.</entry>
+  <entry index="6200356">a heavy battle axe with a blood groove that extends all the way to the handle.</entry>
+  <entry index="6200357">a light flail with dozens of serrated spikes, seemingly to create the most bloodshed possible.</entry>
+  <entry index="6200358">a pike with a wicked piercing tip and a blood groove flowing all the way down to the handle.</entry>
+  <entry index="6200359">a rapier with a triple pointed tip and three blood grooves all the way to the handle.</entry>
+  <entry index="6200360">a sword with a point designed for open wounds and two slashing edges. A blood groove flows down to the handle.</entry>
+  <entry index="6200361">a massive two handed spiked warhammer.</entry>
+  <entry index="6200362">a clear glass bottle.</entry>
+  <entry index="6200363">a clear glass bottle.</entry>
+  <entry index="6200364">a clear glass bottle.</entry>
+  <entry index="6200365">a clear glass bottle.</entry>
+  <entry index="6200366">a clear glass bottle.</entry>
+  <entry index="6200367">a clear glass bottle.</entry>
+  <entry index="6200368">a clear glass bottle.</entry>
+  <entry index="6200369">a clear glass bottle.</entry>
+  <entry index="6200370">a clear glass bottle.</entry>
+  <entry index="6200371">a clear glass bottle.</entry>
+  <entry index="6200372">a clear glass bottle.</entry>
+  <entry index="6200373">a clear glass bottle.</entry>
+  <entry index="6200374">a clear glass bottle.</entry>
+  <entry index="6200375">a clear glass bottle.</entry>
 
   <entry index="6250001">The armor seems quite ordinary.</entry>
   <entry index="6250002">The combat adds for this weapon are +5.</entry>
--- a/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Blood Weapons/BloodWeapons.cs
+++ b/Source/Kesmai.Server/Game/Items/Equipment/Weapons/Blood Weapons/BloodWeapons.cs
@@ -1,0 +1,680 @@
+using System.Collections.Generic;
+using System.IO;
+using Kesmai.Server.Game;
+using Kesmai.Server.Network;
+
+namespace Kesmai.Server.Items
+{
+	public abstract partial class BloodWeapon : MeleeWeapon, ITreasure
+	{
+		/// <inheritdoc />
+		public override int LabelNumber => 6000079;
+
+		/// <inheritdoc />
+		public override ShieldPenetration Penetration => ShieldPenetration.Medium;
+
+		/// <inheritdoc />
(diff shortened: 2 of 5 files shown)


Content/Localization/Enu.xml
```diff
@@ -691,6 +692,28 @@
...
   <entry index="6200353">a large gnarled staff with a ram skull attached.</entry>
+  <entry index="6200354">a scythe with a blood groove leading all the way down to the handle.</entry>
+  <entry index="6200355">a sharp dagger with a blood groove that extends all the way to the handle.</entry>
+  <entry index="6200356">a heavy battle axe with a blood groove that extends all the way to the handle.</entry>
+  <entry index="6200357">a light flail with dozens of serrated spikes, seemingly to create the most bloodshed possible.</entry>
+  <entry index="6200358">a pike with a wicked piercing tip and a blood groove flowing all the way down to the handle.</entry>
+  <entry index="6200359">a rapier with a triple pointed tip and three blood grooves all the way to the handle.</entry>
+  <entry index="6200360">a sword with a point designed for open wounds and two slashing edges. A blood groove flows down to the handle.</entry>
+  <entry index="6200361">a massive two handed spiked warhammer.</entry>
+  <entry index="6200362">a clear glass bottle.</entry>
```
Chilly021, line 703: I would remove Superfluous entries. I didn't know if this is a place holder?

Chilly021 (changes requested):

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Blood Weapons/BloodWeapons.cs
```diff
@@ -0,0 +1,680 @@
...
+		public override int Weight => 1800;
+
+		/// <inherit />
+		public override Skill Skill => Skill.Flail;
+
+		/// <inheritdoc />
+		public override int MinimumDamage => 1;
+
+		/// <inherit />
+		public override Skill Skill => Skill.Flail;
```
sproatba, line 294: double skill override here

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Blood Weapons/BloodWeapons.cs
```diff
@@ -0,0 +1,680 @@
...
+		public override int MaximumDamage => 14;
+
+		/// <inheritdoc />
+		public override int BaseArmorBonus => 4;
+
+		/// <inheritdoc />
+		public override int BaseAttackBonus => 5;
+
+		/// <inheritdoc />
+		public override int Category => 1;
```
sproatba, line 383: halberd has a different category, should be 15

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Blood Weapons/BloodWeapons.cs
```diff
@@ -0,0 +1,680 @@
...
+		public override uint BasePrice => 10000;
+
+		/// <inheritdoc />
+		public override int Weight => 5000;
+
+		/// <inheritdoc />
+		public override int Category => 1;
+
+		/// <inheritdoc />
+		public override Skill Skill => Skill.Greatsword;
```
sproatba, line 613: says BloodMace but uses greatsword and category 1, is this a 2h mace?

sproatba: flags don't indicate 2h if so

sproatba (changes requested):
I'll let others weigh in, technically not an issue, but I find it much easier to have 1 class per file than a bunch in one big one.  We have mostly that, but there are a few big ones in there currently.     I'll have to look at the bottles section later.

SewerCzar:
Resolved angry's well caught items. I didn't realize Halberd had its own category - must be for items that cant be belted.

As for the large class file I thought it would be best so I could reference the base BloodWeapon class right away in one CS file. I know Lux has a lot of big CS files, plus they're for a very specific set of goals. If we have to edit them all we can do them in the same file now. In my head it makes sense but it's hard to give a specific example of why it is. Lets say we want to add consumabletype_ edits to all of them, it's easier? I dunno. Thanks for the look over

Chilly021:
> I'll let others weigh in, technically not an issue, but I find it much easier to have 1 class per file than a bunch in one big one. We have mostly that, but there are a few big ones in there currently. I'll have to look at the bottles section later.

Yea, creates less confusion in the design pattern. Typically you can have generic classes that handle this, but I'm not going into this now. Easier just to separate the files and create a separate file for each weapon/bottle.

zrikzlok:
I can see the benefit either way of having the separate or not. Mass update or similar things since these are all bloodlands related would potentially make life easier if we notice a problem that needs to be changed on all.

zrikzlok:
I can’t seem to find the comment where chilly mentioned the same names for multiple enums. Place holder for now , suppose we could add a number or letter to differentiate them , but need to come up with creative text for them that we weren’t quite ready to finalize.

zrikzlok:
As long as there are no technical issues observed I think this is ok to see if it functionally works without issue. We can debate separation post test , this way if there are issues in various files at least it will only require minimal file edits

zrikzlok (approved):

Chilly021:
Still need the class changes for the alchemist. Compiler will freak out otherwise.

zrikzlok (approved):

Chilly021:
I'll check back after kids are to bed. Class/constructor look good! Just need to check for syntax errors and should be able to approve tonight or early tomorrow.

sproatba (approved):

Chilly021 (approved):
Looks like most things covered in the update.

sproatba:
I don't see anything right off that will cause things to break.  It would probably be good to have a conversation on AR/DR progression on weapons and where we want to target them initially for the land.  It looks like all these are pretty much kitchen sink items with bg/silver good offence and defense.   These probably superscede almost all bg weapons except maybe raxe/rhammer at first glance.  Not necessarily a bad thing, but we want some amount of variance on what they do.

SewerCzar:
> I don't see anything right off that will cause things to break. It would probably be good to have a conversation on AR/DR progression on weapons and where we want to target them initially for the land. It looks like all these are pretty much kitchen sink items with bg/silver good offence and defense. These probably superscede almost all bg weapons except maybe raxe/rhammer at first glance. Not necessarily a bad thing, but we want some amount of variance on what they do.

Yeah all these items aren't particularly powerful - I'm expecting the itemization balance pass to turn things on its head a bit so it's just +1 +1 over BG weapons... except all the fucking outliers but that's just LOK.

It's intended for a system of weapons / drops / crafting that Z and I are formulating - nothing set in stone but this will enable us to properly tune how things drop, where they drop, and what we want them to craft after they drop, if that makes sense.

System we've been discussing for a while is bloodlands weapons harvest blood from monsters when killing them, which are then used to create bloodlands items, possible upgrades, valuable gems depending on how you combine them, etc. They're created at the Alchemist NPC.

Nothing in here is in any kind of final form but it will allow us to start creating blood drop scripts and filling in the alchemists craft table, which is going to be super complicated. Someone with programming experience might want to do that.

I'm stepping back from the item balance here for sure. I feel like I just said a bunch of stuff more than once so for clarifications hit me up in Dev channel.

PR #1651, opened 2022-05-06
Fixed grove heart

SewerCzar:
Turned grove from sloppy shield shape to actual heart shape

Evened out spawns with a <.20 density and 25-30 minute respawn (pre-adjustment)

Removed the random cat demons that prevented the zone from being particularly farmable.

Increased gems to be equivalent to axe undead - difficulty should be similar.

Revamped escort quest - just one cat demon resides in the aorta passage, killing it will despawn the obstacles and begin the escort but the escort is farther now. 

Upon finish of the escort quest a boss is released from behind the blocked pulmonary veins.

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -1229,6 +1229,31 @@
 //                var tile7 = Segment.FindTile(34, 51, 236);
 //                tile7.Add(new Obstruction(266, false));
                 new OakGroveBow().Move(Location, true, Segment);
+                
+                Emote($"the heart begins absorbing fresh nutrients from the ground, but it shivers as a furious meow is heard in the distance.");
+                
+                EmitSound(228, 3, 9);
+                
+                EmitSound(228, 3, 9);
+                
+                EmitSound(228, 3, 9);
+                
+                var tile = source.Segment.FindTile(72, 11, 236);
+    
+                if (tile != null && tile.ContainsComponent<Obstruction>())
+                {
+                    var component = tile.GetComponent<Obstruction>();
+                    tile.Remove(component);
+                }
+                
+                var tile2 = source.Segment.FindTile(77, 17, 236);
+            
+                if (tile2 != null && tile2.ContainsComponent<Obstruction>())
+                {
+                    var component = tile2.GetComponent<Obstruction>();
+                    tile2.Remove(component);
+                }
+                
                 Timer.DelayCall(TimeSpan.FromSeconds(0.2), () => Delete());
             }
             EmitSound(220, 3, 9);
@@ -1478,6 +1503,7 @@
             }
         }
     }
+    
 ]]></block>
     <block><![CDATA[]]></block>
   </script>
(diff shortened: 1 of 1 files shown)


zrikzlok:
Pretty dang cool looking map I must say, have only minimally played in grove but this looks interesting.

zrikzlok (approved):

Leolus (dismissed):

Chilly021 (changes requested):
Naming of variables is not following conventions already used prior. 
image

Additionally there are many changes not listed here. 

* Stairs up to Teleporter to Sea Serpent 
* Moved Sogmallon to heart necro spawner (assuming this is the update: just one cat demon resides in the aorta passage, killing it will despawn the obstacles and begin the escort but the escort is farther now.)
image

* Aorta is brand new and can be entered through two drops
image

SewerCzar:
> I don't see anything technically wrong here. Balancing might be an issue as mentioned. I'll be interested to see how all the area strips will work out for mob flow. I think fix up what chilly pointed out and we are good

image

Already exists

image


No idea what this means? Thats the total number of packs? Density is sub .2

I have NO clue why every single spawn is posted but densities are sub .1 on each individual spawner and sub .1 on the entire spawner.

Spawner being done this way is best practice now.

image

Grouping them together as GroveHeart is also best practice? 

I'm really confused by every single spawn being posted here. We can see density, we can see location, we can see they're within density norms and respawn slightly slower?

And the aorta section is just a part of the new map it isn't notable at all?

And the Boss NPC in the pond won't leave that section as he's tethered by his spawn.

Honestly just confused here

SewerCzar:
> I guess we can fix it for you? Please let us know if you need us to commit the changes.

I responded to all of your weird post. Fix what? All I see in that post is a bunch of screenshots of spawns that don't need fixing, and a set of stairs that already exists. 

Number of monsters is equivalent to axe - I have no idea why thats mentioned - see above about density being like Axe.

Spawns are split up in a way thats... entirely normal? It's spawns with multiple boundaries? This works in the game in every single map, as it's currently in use.

Stairs to the grove heart already exist - see above

Boss monster spawning in pool - he will never leave that one section which won't be used for hunting at all, and is outside for testing.

What needed fixing? This is such a bizarre "request"

Chilly021 (changes requested):
Please correct naming conventions and we should be good here. A few of us spoke.

Chilly021 (approved):

SewerCzar (commented):
Naming conventions have been settled on - lowercase one string.

Previous prove monsters have been sectioned off with a groveEntrance label instead of just grove.

sproatba (approved):

Chilly021:
> Naming conventions have been settled on - lowercase one string.
> 
> Previous prove monsters have been sectioned off with a groveEntrance label instead of just grove.

thank you, sorry to be a pain.

zrikzlok (approved):

PR #1657, opened 2022-05-10
Add Midgits work to Kes

Chilly021:
* Add a new region and add Midgit's map to this. 
* Not populated - and no way to access the area quite yet. 

image

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<segment name="Kesmai" version="0.94.0.0">
+<segment name="Kesmai" version="0.95.0.0">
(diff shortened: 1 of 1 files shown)


sproatba:
Missing some floors here
image
image
I'm not entirely up to speed on the stairs issue, but I think there is some kind of problem with them functioning properly if they have a floor under them or something.  They were dealing with it in blood lands recently, most stairs just don't have anything on the hex with them.
image
same thing with the -4 steps

Everything looks good to me as far as merging the map files goes, everything is untouched outside of adding this region and the steps on -4.

Chilly021:
@sproatba  should be good now.

sproatba (approved):

Fireal1983 (approved):

Leolus (approved):

PR #1658, opened 2022-05-11
Fix Dino Boss Lair and add more density in caves for forest

Chilly021:
1) Dino boss lair didn't have exclusion area in it. So it was getting crits from forestTop
2) Caves were super sparse in density. Upped the density and adjusted max capacity


Segments/Bloodlands.mapproj
```diff
@@ -46719,7 +46719,7 @@
       </tile>
       <tile x="18" y="-67">
         <component type="FloorComponent">
-          <ground>12</ground>
```
sproatba, line 46722: this looks like the y got typed in the wrong field maybe?

sproatba (changes requested):

Segments/Bloodlands.mapproj
Chilly021: Yep, for whatever reason when you type in exclusion area with a component up, it types in component as well? 

image

Chilly021:
@sproatba  back up for you.

sproatba (approved):

zrikzlok (approved):

SewerCzar (approved):

PR #1664, opened 2022-05-12
Kesmai -5: Flip flop the stairs to allow UP

Chilly021:
Helping out Midgit one more time before he takes on the population piece. 

Currently this:

image

is not behaving with this: 

image

Issue: User can not type up on the stairs below. They can type down on the stairs above. I wouldn't imagine the stairs to correspond to each other, but attempting one small fix to see if it works. 

Attempting to flip flop the stairs down to fix it.

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -76936,7 +76936,7 @@
           <destinationX>31</destinationX>
           <destinationY>37</destinationY>
           <destinationRegion>71</destinationRegion>
-          <teleporterId>126</teleporterId>
+          <teleporterId>123</teleporterId>
         </component>
       </tile>
       <tile x="15" y="2">


zrikzlok:
If this doesn't solve the issue, as frustrating as it is, for unknown reasons I've had to move multiple stairs by 1 tile in bloodlands to resolve up/down issues.

sproatba:
just a theory, but are the stairs using elevation to decide instead of the action set on them?  stairs are set to action 6 down on 123 and 126
image

sproatba (approved):

Chilly021:
> If this doesn't solve the issue, as frustrating as it is, for unknown reasons I've had to move multiple stairs by 1 tile in bloodlands to resolve up/down issues.

Sure thing! if you want to throw a commit. I was just trying to figure it out myself. Only noticeable difference I saw at first was the order of the facing of the stairs. I'm waiting on this for crit implementation.

Chilly021:
Fixed with https://github.com/jkachhad/Stormhalter/pull/1676

PR #1665, opened 2022-05-12
Graphics tweaks. (Shadow opacity, Wall Edges, corpse coordinates, Firebowl temporary fix.)

Tenscore:


Tenscore:
Updated Terrain.xml to include action attributes, and sorted sprite-orders.

sproatba (approved):
I'm unsure of the action="0" on everything but that might be ok.

I'm also unsure of setting the action on the stair decorations.  Again might be ok, or it might be giving the up/down command to a tile with no destination logic.  Might need to run this by someone with a little more knowledge.  Other than that I don't see anything out of sorts.

Tenscore:
as they effectively cover the stairs I err'ed on the side of caution - so if it breaks stuff, it can be easily fixed.
Also, unless default is null, i figured 0 would be ok.  Would rather have them in, then out so no one has to do this again.

Chilly021 (approved):

zrikzlok (approved):

PR #1666, opened 2022-05-12
Axe Naming Convention Fix

SewerCzar:
Was recently brought to attention that axe is using a nonstandard naming system. This has been fixed

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -111231,7 +111231,7 @@
         <exclusion left="0" top="0" right="0" bottom="0" />
       </bounds>
     </spawn>
-    <spawn type="RegionSpawner" name="BossMama">
+    <spawn type="RegionSpawner" name="bossMama">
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

sproatba (approved):

Leolus (approved):

Tenscore:
wait... so should we have instructions on "Naming conventions" for world-makers?
Sure sucks to have to go back and redo everything just because no one told us in advance.

sproatba:
We currently don't really have any official standards, but I'd like to put some together.  For now we are trying to remain consistent with what already exists until we get something together.   Right now it's just tribal knowledge on what works and what doesn't and why we do certain things.  It's creating a lot of work for some trying to watch for and clean up when people are changing things without understanding the implications of what they are doing.  This one was probably fine the way it was, and I think most will find the old format more friendly.  What it is changing to however matches what is going on in most of the other segments.  Either way need to make a decision on which is better and stick to it.

PR #1670, opened 2022-05-19
Add Surface Level Annwn Crit Population

Chilly021:


jkachhad:
If possible, can we do this bosses first?  This would a allow parties to venture to the area and test often before meeting barriers.

Chilly021:
Sure!

PR #1671, opened 2022-05-19
update for guardian test

TimPickup:
1) he wasn't dying correctly I think this was due to his onDeath code creating and moving the item. I now specify more detail about the segment to move the item to.
2) the guardian correctly Wakes when you move on his square but then is loose if you flee or die. There is now a timer that waits 2 mins since he last cast a spell on someone (aka saw someone) after which he is reset / teleported back to his start position and make him not move again.
3) The code where I try to stop more than one player in the lair at once did not work so I'm trying a cleaner version where we check all players subregion.. if it matches the guardians lair then we refuse entry.
4) Also removing a test door I put in the Test area

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -488,6 +488,42 @@
             Movement = 3;
         }
 
+        public override void OnSpellTarget(Target target, MobileEntity combatant)
+        {
+            //Start a timer - if we don't restart this timer again then we have lost sight of them for a long time and we reset the guardian
+            StartResetTimer();
+            base.OnSpellTarget(target, combatant);
+        }
+
+        private Timer _resetTimer = null;
+        
+        private void StartResetTimer()
+        {
+            if (_resetTimer != null)
+                _resetTimer.Stop();
+                
+            //if the guardian doesnt see someone for 120 seconds then reset him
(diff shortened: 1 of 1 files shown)


SewerCzar (approved):

Chilly021 (approved):

yarrim (approved):
This looks good to me. I'm a little concerned about the Portal Entity's OnIncomingPlayer code where it looks through all mobs in the World. Can we figure out how to scope that better to only scan the segment or a region, or trigger a manual spawn when a player has the right item? As we add more content, I can see that evaluation over all mobs becoming costly.

TimPickup:
Thanks everyone for looking over it. I appreciate some code is unknown as such so always a risk.

Regarding the "all mobs in the World" that was actually a optimisation/cleaner code I think than the other way I was trying to do it suggested by Angry.

Its not "all mobs" its all Players on the current Segment/Map in that facet.. so not many people to loop over. Its the same code used to announce spawned bosses to players. Then it works out which segment each player is in like a recall ring does (but recall ring is checking if recall is allowed) so I think all in all, its fairly neat way to do it. 

Being able to just query users in a segment would be ideal though but I don't know of any examples of such code.

PR #1674, opened 2022-05-21
leng - fixing holes in spawn area

TimPickup:
There are various holes is spawn areas that cause mobs to get Stuck in 1 area unable to move around as intended. There were also some safe spots where mobs would never move. Note: Only Spawn areas are tweaked - nothing else

I have tried to make it so at worse there is never more than 2 spaces away from a spawn on appropriate areas.

I will detail the changes bellow: Long post but trying to make it visually very easy to approve.

**Maus**
From:
image
To:
image

From:
image
To:
image

From:
image
To:
image

From: **Important one - mobs got stuck in the crypt and couldn't get out**
image
To:
image

From: This allows movement through the whole crypt area
image
To:
image

From:
image
To:
image

**Shid**
From:
image
To:
image

From: **important - mobs got stuck in SW room unable to roam until disturbed**
image
To:
image

**undead** note: NW spawn was fine but NE and S spawns were very odd and not covering the areas correctly
From : notable change..
image
To:
image

From: This was a really random area to call NE. This makes a lot more sense
image
To:
image

From: 
image
To:
image

UD as a whole overview .. 
From:
image
To:
image

Thats all folks

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -81317,7 +81317,6 @@
       <entry entity="level13Troll" size="3" minimum="1" maximum="10" />
       <bounds region="18">
         <inclusion left="16" top="27" right="43" bottom="38" />
-        <inclusion left="0" top="0" right="0" bottom="0" />
         <exclusion left="0" top="0" right="0" bottom="0" />
       </bounds>
     </spawn>
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -80243,6 +80243,7 @@
       <entry entity="level9Wraith" size="1" minimum="1" maximum="1" />
       <bounds region="10">
         <inclusion left="2" top="2" right="32" bottom="5" />
+        <inclusion left="25" top="6" right="33" bottom="7" />
         <exclusion left="0" top="0" right="0" bottom="0" />
       </bounds>
     </spawn>
@@ -80272,6 +80273,8 @@
       <entry entity="level9HobgoblinMage" size="1" minimum="1" maximum="1" />
       <bounds region="10">
         <inclusion left="7" top="6" right="11" bottom="33" />
+        <inclusion left="3" top="22" right="8" bottom="37" />
+        <inclusion left="12" top="5" right="13" bottom="7" />
         <exclusion left="0" top="0" right="0" bottom="0" />
       </bounds>
     </spawn>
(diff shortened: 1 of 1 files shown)


SewerCzar:
Do we want to continue the whole thing where there are 0 safe spots anywhere near hunting grounds? Because I don't think anyone likes that at this point. Notably climbing up Leng Undead rope to a potential zoo seems fucked. And speaking of Leng Undead the point of the weird spawns was "Leng Undead had secret rooms specifically with zoos in them." I'm fine extending them thats how I originally did it but I think climb ups shouldn't necessarily be full of monsters in zones where you can get zooted.

The ninjas in Shid Lair are invisible but extending them is fine.

TimPickup:
The NE UD currently mostly misses the secret room.. they can spawn in 1 of 3 areas basically as shown here:
image

Thats 10 groups of mobs spawning in a very small area (same amount as the whole NW area)

The South area has some secret doors but no actually totally cut off by secret doors:
image

The area up from leng town is unlikely to zoo up - its only included in the whole UD spawn and they have free roam to leave that area so no reason at all it would zoo up any more than any other stairs or rope. There is a easy jump to escape and mobs cant follow you back down. Even with the potential of some creatures there, its still the safest way into any UD area that I can think of.

Whats the thinking behind having safe areas for people to cast / thieves attack from? Maybe I'm missing something.

Some of the above are outright bugs and some are safe spots and some are just my thoughts so happy to discuss all and remove any if there is consensus that they are bad

TimPickup:
somehow missed the waterfall/duck area out. There will still be that "almost" safe zone next to the princess with only 1 centaur but I'm not going to try to change that.
Was:
image
To:
image

Chilly021:
You might want to split up spawns in Leng ud. Just from a rough comparison of what I just added in oak serp, some rooms will still spawn crits more-so. I had to separate it into quadrants and make sure to distribute evenly there. If you have one big area for spawn location, you can have some weird stuff happen. We still get issues with certain areas across the board.

TimPickup:
As just discussed with you Chilly - I haven't changes the main spawns of UD much - it was always 1 big area. I changes the smaller areas a bit. As another PR we could phase out the big spawn into the updated NW, NE and S spawns but I would prefer not to add that compilation into this PR

sproatba (approved):

Chilly021 (approved):

zrikzlok (approved):

PR #1677, opened 2022-05-22
Oak -1 polish

TimPickup:
**Overall Goal:**
_Current difficulty:_
-1 oak  = -2 Kes
-2 oak =  between -3 and -4 Kes probably similar to Axe cave.
-3 serp = just harder than -4 kes 
-4 TT = slightly harder than serp
-5 UD is a lot harder than TT

_Im thinking....._
make -1 more useful .. about difficulty of axe cave/-3 kes
-2 very slightly harder (between new oak -1 and -3)
-3 leave
-4 make a little harder between serp and ud difficulty
-5 leave

**What I have done:**
I have tired to make -1 Oak more useful to players that could reach it. On par with Axe caves, but different due to different mobs.

I have deliberately left a path though that has a chance of slightly less density so as to try to reduce impact on Konran quest. basically the E and W spawns avoid this path (but do over see it) and there is a whole area spawner what does cover the path.

This is going to be something we need to feel if its right.

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -110887,40 +110887,38 @@
         <block><![CDATA[]]></block>
       </script>
     </entity>
-    <entity name="level3Rat">
+    <entity name="level6Rat">
       <script name="OnSpawn" enabled="true">
         <block><![CDATA[]]></block>
         <block><![CDATA[
 	var rat = new BlackRat()
     {
-        MaxHealth = 18, Health = 18,
-        BaseDodge = 14,
-
-        Experience = 482,
-
+        MaxHealth = 30, Health = 30,
+        BaseDodge = 18,
+        HideDetection = 18,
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

TimPickup:
Hopefully this makes it easier to see what Im doing..

**Some stats:**
Taken the mob count from about 32 to 66. Density of about 0.28 .. about the same as Axe caves.

The threats and rewards are very on par with Axe cave as seen here:
**Axe Cave:**
image
**Oak spider**
image

Spawn areas:
There is a spawner covering the whole map and then overlay ones to build extra density and  contain some structure while leaving an element of random still.
image
Although as you can see the top up areas leave a slight path through for Konran quest.. but you may pick up a few mobs when they see you passing.
image
image

SewerCzar (approved):

sproatba (approved):

PR #1679, opened 2022-05-23
moving the test guardian code to live

TimPickup:
All tested now in the Kes Test area - now moving to UK.

As discussed previously but didnt work -

This will now consume the amulet on entry.

It also restricts the area to only allow one player at a time as a personal challenge.

Guardian throws fairly frequent WWs but they are just about balmable for an appropriate level player. All in all its not supposed to be a very hard challenge as they have already "paid" its more of a check that a very low level isn't being gifted a hummer.

I have aso removed all the old test code from Kes

### DIFF
--- a/Segments/Kesmai.mapproj
+++ b/Segments/Kesmai.mapproj
@@ -440,111 +440,6 @@
     
     /*************** TEST area code section  ****************/
     
-    public class TESTHummingPortal : CreatureEntity
-    {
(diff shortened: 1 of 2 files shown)


Chilly021:
Good job testing this and getting it out!

Chilly021:
Uno mas on this. If we can add it to UK for now but not remove from Kes yet

PR #1682, opened 2022-05-24
Oak 2 and 1 tweak

TimPickup:
so I had to re-do my -2 update! due to conflicts caused by the -1 update.

anyway.. I have also put in some tweaks to -1 after testing it. -1 had too many snakes and scorps at once, so fixed that. They and spiders also did too much poison damage for the difficulty so I tweaked that. I also slightly increased their exp as they drop nothing and have a higher risk factor due to poison so reward via slightly higher exp a bit like we do for archers and MUs. oh and I also tweaked the spawn areas so maybe less zoo as soon as you go down.

-2 I did same as before except I decided to tweak the spawn area to make it headier density in the East (north and south) leaving a slightly fairer route for Konran. I did also add in an extra secret door for 2 reasons.. I want to encourage going through that secret room as it has its own mini spawn area now, and also it is easier access to the East area for hunting + access to the rope no one uses to the Eastern Ledge.. again maybe an area we could make useful.

image

image

Let me know any questions.

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -41519,25 +41519,25 @@
         </component>
       </tile>
       <tile x="15" y="15">
-        <component type="StaticComponent">
-          <static>179</static>
+        <component type="FloorComponent">
+          <ground>179</ground>
         </component>
         <component type="WallComponent">
           <wall>158</wall>
           <destroyed>161</destroyed>
           <ruins>170</ruins>
-          <indestructible>true</indestructible>
         </component>
       </tile>
       <tile x="16" y="15">
-        <component type="StaticComponent">
-          <static>179</static>
+        <component type="FloorComponent">
+          <ground>179</ground>
         </component>
-        <component type="WallComponent">
-          <wall>157</wall>
-          <destroyed>160</destroyed>
-          <ruins>170</ruins>
-          <indestructible>true</indestructible>
+        <component type="DoorComponent">
+          <openId>74</openId>
+          <closedId>62</closedId>
+          <secretId>157</secretId>
+          <destroyedId>82</destroyedId>
+          <isSecret>true</isSecret>
         </component>
       </tile>
       <tile x="17" y="15">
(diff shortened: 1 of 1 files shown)


Chilly021 (approved):

sproatba:
spawns seem way off here, the maximum on them is very low in comparison to how many mobs could potentially spawn in each region.  The density value isn't based on the maximum but rather the entire collections maximum.  I'm also unsure we want to completely prevent the secret room from becoming a zoo with the exclusions.  Getting mobbed opening a door is part of the fun.  If you know exactly what is going to be there every time that's fine, but I'd rather see that in a more quest/lair situation than just a normal hunting zone.  I'm open to giving it a shot and see how it plays out, but I have a feeling there just isn't going to be enough mobs in there.  I also see the gems drop rate reduced in one spot in there, was there too many dropping?

TimPickup:
@sproatba The density is based on the Span Maximum. Really it should also take into account if the Creature maximums are under that.. but in my case they are over so not an issue. The server would in the case bellow spawn 3 creatures out of the list. 1 to 2 hobs and the other 1-2 mobs as other creatures. IMO the spawns are right and I would like to try it. Note you have to move off and back a spawn to get it to recalculate the density. Also note, I also see creature minimum values often that go Over the overall region Maximum which makes no sense at all.. having low minimums will mean more random of the creatures encountered and only guarantee certain specific mobs I really want to always be there. 

oh and also - note there are some overlapping spawns.

See examples changing Overall maximum  

As I defined:
image

Changing the Spawn Maximum:
image

Regarding the secret room.. could not exclude it. wanted to try it as is as it will guarantee a small random zoo in there. Always 3 groups yes but they could be 2 sentry and 1 hob or could be 2 hobs and a gob, or a spec, trapper and hob etc so varied (always 1 hob to go with the hobgoblin cave theme)

sproatba (approved):

SlysterLOK:
I never can get that rope to work. Am I the only one? lol

Looking forward to trying it out.

Chilly021:
Let me bump this.

Fireal1983 (approved):

PR #1683, opened 2022-05-25
Kes UD Spawn Update

SlysterLOK:
Increased the spawn as it was pretty barren felt like.

The 2 mini bosses were also not spawning so I just through them in their own region spawn instead of doing a sub region for the lair since they will roam the whole area anyway.

Tweaked the map a bit as well.


SlysterLOK:
Oops! No problem. Thanks for pointing that out again. Slipped my mind

Chilly021:
Just in case, see dev channel. If you could change this back to X = 1 and Y = 1 
image

SlysterLOK:
Done

Chilly021 (approved):

sproatba:
image
May need to adjust the no recall zone or add some spawns in the new hallway or something.

sproatba:
image
are the ydnac -1 areas supposed to still be there?   If so should they be renamed to the kes 3.5?

sproatba:
image
spawns here can end up where no player can get to, and potentially can walk off the map or something else odd.  I think we want an exclusion here.

sproatba:
image
same over here

sproatba:
image
looks like these two spawns regions were orphaned with a region change.

SlysterLOK:
Doesn't look like you have the newest push. I can tell because the solo mummy spawns are gone and the segment has been renamed already.

I have the water as an exception but haven't committed it yet.

I don't think the hall needs to be no recall or have mobs since the area south and West aren't no recall nor does it have mobs. If anything I would remove those stairs and hall.

SlysterLOK:
I went back and look and you were right! Somehow those spawn regions were still active. I am positive I deleted those before even starting the expansion work. Oh well. Taken care of now.

SlysterLOK:
Ok. Updated. Maybe this one will suffice.

SlysterLOK:
I cannot for the life of me find the issue. I've deleted rewrote each change and still cannot get to the bottom of it.

SlysterLOK:
Ok I finally got it. I had to roll back a bit but I think I got it all ready to go.

sproatba:
image
image
I don't think we can use 4.5 for skill level.  If it does compile I think it just does lvl 4 damage.

sproatba (approved):

Chilly021 (approved):

zrikzlok (approved):

PR #1685, opened 2022-05-31
Fix -5 Kes Stairs and add an altitude to the region

Chilly021:
Add Altitude and move stairs a bit to adjust for stairs fix


sproatba (approved):

Leolus (approved):

Chilly021:
Resolved in https://github.com/jkachhad/Stormhalter/pull/1693

PR #1703, opened 2022-07-10
Kesmai -5 - Exclusion area for Generic Spawner. No extra crits in Blood Ghoul Area

Chilly021:
Read Title - Two small exclusion areas.


Chilly021:
Conclicts with other PR that Midgit put out there

PR #1711, opened 2022-07-13
Wall fixes for opening doors

Chilly021:
Wall fixes for opening doors 

- This an attempt, I think it needs a recompile too. Not too sure what will happen with existing walls and their components.


Chilly021:
Tenser went back and added this today. So it should cover both. Let's hope! https://github.com/jkachhad/Stormhalter/pull/1712

PR #1715, opened 2022-07-17
Add Nightvision to Bears / Bats / Wolves

Chilly021:


Leolus (approved):

zrikzlok:
Should all animals have night vision though? this will make surface kes and other places a bit more difficult than they probably should be I'd think.

Tenscore:
> Should all animals have night vision though? this will make surface kes and other places a bit more difficult than they probably should be I'd think.

"...even in the absence of vision, they'd still have their 'smell'." <--- OG view that animals smell-sense and eye-sense is better than humanoids.

Surface kes animals aren't too terrible if you don't let them stack up too much.
tbh, daytime is worse bc of archers.

sproatba (approved):

Chilly021:
Suspending conversations for this one. Overall I think we either need to create a monster level + NV protect or overall get on the same page with our resists. This was a conversation that took place last week in regards to what animals need NV in the end. 

The decision was to roll with wolves and bears. So as to honor that for now, we will add it there. Boars / sallies / everything else though will stay on the same page with no NV.

sproatba (approved):

PR #1716, opened 2022-07-17
Add NV to all base classes for Undead (except a few). Interface was exposed for IUndead but no Undead. 

Chilly021:
Sources: 
/* https://www.themonstersknow.com/undead-tactics-specters-wights-wraiths/ - Nightvision add*/
/* https://www.themonstersknow.com/undead-tactics-ghouls-ghasts/ - Nightvision add */
/* https://www.dragonsfoot.org/forums/viewtopic.php?t=59544 - Nightvision add*/

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Wraith.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Wraith.cs
@@ -14,6 +14,8 @@ namespace Kesmai.Server.Game
 			Body = 50;
 
 			Alignment = Alignment.Chaotic;
+			/* https://www.themonstersknow.com/undead-tactics-specters-wights-wraiths/ */
+			AddStatus(new NightVisionStatus(this));
 		}
 
 		public override void OnSpawn()
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
@@ -15,6 +15,8 @@ namespace Kesmai.Server.Game
 			Body = 12;
 
 			Alignment = Alignment.Chaotic;
+			/* https://www.themonstersknow.com/banshee-tactics/ - Nightvision add*/
+			AddStatus(new NightVisionStatus(this));
 		}
 
 		/// <inheritdoc/>
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Ghoul.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Ghoul.cs
@@ -15,6 +15,8 @@ namespace Kesmai.Server.Game
 			Body = 77;
 
 			Alignment = Alignment.Chaotic;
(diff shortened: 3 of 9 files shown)


Leolus (approved):

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Wraith.cs
```diff
@@ -14,6 +14,8 @@ public Wraith()
 			Body = 50;
 
 			Alignment = Alignment.Chaotic;
+			/* https://www.themonstersknow.com/undead-tactics-specters-wights-wraiths/ */
```
sproatba, line 17: what is this comment on all of these for?

sproatba (approved):

zrikzlok:
I get it on the dnd rule aspects. But does this line up with OG or lands aspects for undead vision ? I mean, in lore, orcs can smell well, should they all see NV and detect hide / life force ?

I feel like this will make kesmai a bit(good amount more) more difficult for all do the undead gaining nv, no ?  

Also can’t you hide from banshees and the like in UK with dark ? Been a while perhaps not, but I know a lot of undead you currently can I suspect ?

Chilly021:
> I get it on the dnd rule aspects. But does this line up with OG or lands aspects for undead vision ? I mean, in lore, orcs can smell well, should they all see NV and detect hide / life force ?
> 
> I feel like this will make kesmai a bit(good amount more) more difficult for all do the undead gaining nv, no ?
> 
> Also can’t you hide from banshees and the like in UK with dark ? Been a while perhaps not, but I know a lot of undead you currently can I suspect ?
> 
> What is the point of dark if animals and undead can see NV. We can hide from orcs/trolls or humanoids for random reasons ?

I am sorry I missed this. Been a busy day. We can definitely adjust it. It's a baseline. I was putting this as a suggestion since right now Oak/Axe/UK are all different. Oak seemed the closest to this. If you look it up all of the crits that are undead have night vision. If we need to change the mechanics of individual, overrides are there for this.

Chilly021:
So we can close this for now, but someone needs to address the inconsistencies between all of the Undead areas. I'm not calling a specific one out, because guess what.. they are all different. 

More or less a decision point. We have Nightvision a bit muddled and inconsistent. It's as clear as mud as to what creatures should and should not have it. 

Axe UD != Oak UD != Leng UD != Kesmai UD != Bloodlands UD != UK UD

Everything is pretty different. If that is intentional, from a bestiary standpoint, it's a bit sloppy from an implementation standpoint for newer people coming in an re-using the same implementation. From a OOP standpoint, it makes sense to inherit down the object itself and just overriding it in the MapProj where you see fit. 

Maybe an Undead somehow gained EyeSight, who knows. But this wasn't supposed to "replace" existing, but just add a base pattern.

Chilly021:
More or less a decision point. We have Nightvision a bit muddled and inconsistent. It's as clear as mud as to what creatures should and should not have it. 

By default Undead usually do not have eyes as a living creature would. Skeletons don't even have eyes as at all. If it makes sense to model everything off of OG, we aren't even doing that. 

Axe UD != Oak UD != Leng UD != Kesmai UD != Bloodlands UD != UK UD

Everything is pretty different. If that is intentional, from a bestiary standpoint, it's a bit sloppy from an implementation standpoint for newer

PR #1718, opened 2022-07-18
Boss Properties - Set Defaults of boss weaknesses and immunities based on Boss Type

Chilly021:
A new idea to standardize properties for Bosses. This will set defaults up front and then the designer can then abstract the properties even more in the map proj file. 

Boss Types = Major, Notable, Minor

@jkachhad  let me know if this would be a good quality of life improvement. Could clean up the mapproj files up a good bit where we are having to keep setting properties.

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/CreatureEntity.BossProps.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/CreatureEntity.BossProps.cs
@@ -1,0 +1,88 @@
+using System.IO;
+
+using Kesmai.Server.Accounting;
+using Kesmai.Server.Engines.Commands;
+using Kesmai.Server.Items;
+using Kesmai.Server.Miscellaneous;
+
+namespace Kesmai.Server.Game
+{
+    public class CreatureEntityBossProps
+    {
+        private CreatureEntity _creatureEntity;
+
+        public CreatureEntityBossProps(CreatureEntity creatureEntity)
+        {
+            _creatureEntity = creatureEntity;
+        }
+
+        /* This needs a look from Lux. 
+                I don't know if the implementation is right due to access of source code is limited. 
+                Overall this is to clean up the Bosses and to establish Default Props using a c# file. 
+
+                To what extent this could be used would be determined by the person implementing the bosses. 
+            */
+        public CreatureEntity SetBossProperties(BossType bossType)
+        {
+            
+            switch(bossType)
+            {
+                case BossType.Minor:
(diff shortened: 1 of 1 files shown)


Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/CreatureEntity.BossProps.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            if (weakness != null)
+            {
+                _creatureEntity.CreatureWeakness = weakness;
+            }
+
+            switch(bossType)
+            {
+                case BossType.Minor:
```
Leolus, line 39: No default case means you're not going to assign anything if something beyond minor major or notable come through here. That's okay if that's the intent but generally default cases are assigned as a catch unexpected.

Chilly021: no worries. Return null I guess or bail out.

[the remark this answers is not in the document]
Chilly021: and done and back to you @Leolus

Chilly021: and done and back to you @Leolus

Leolus (approved):

sproatba:
I'd like to see something like this if possible,  but I think there are just too many combinations to make it a reality.   If you look at what hits each boss in other versions, and even this one, they are all too different I think to make something like this work.

Chilly021:
So to catch you up, basically we had properties set on CreatureImmunity and CreatureWeaknesses that have been missed by myself and others. I want to see if we can standardize this somehow so at least Major Bosses don't get missed. I could catalogue these, but trying to stay out of parsing xml right now for my sanity sake. 

Otherwise the following properties is left up to copy and paste mistake or up to human error. I figured by creating something for a "Base" of properties people can just extend or override them if they are doing something strictly custom. 

Having to maintain this for each Major boss seems to be a bit redundant and messy. 

   Immunity = CreatureImmunity.Piercing | CreatureImmunity.Slashing | CreatureImmunity.Bashing |
                   CreatureImmunity.Projectile | CreatureImmunity.Magic | CreatureImmunity.Poison,
        Weakness = CreatureWeakness.BlueGlowing | CreatureWeakness.DeathSpell | CreatureWeakness.IceSpearSpell,

PR #1719, opened 2022-07-18
Fixed undead monsters according to monster manual

SewerCzar:
Source: https://www.dndbeyond.com/monsters

We can't access damage reduction for some types so had to add some permanent resists to make it match. It's closer this way to manual accurate.

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
@@ -6,6 +6,13 @@ namespace Kesmai.Server.Game
 {
 	public partial class Banshee : CreatureEntity, IUndead
 	{
+
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Piercing;
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Slashing;
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Bashing;
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Projectile;
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Poison;
+		public override CreatureWeakness Weakness { get; set; } = CreatureWeakness.Silver;
 		/// <summary>
 		/// Initializes a new instance of the <see cref="Banshee"/> class.
 		/// </summary>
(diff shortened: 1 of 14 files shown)


Chilly021:
Up for conversation, but I see this PR as sarcasm. Hide 100.. and so on. If you think Night Vision does not need to be on Undead, maybe we need to talk about this, rather.

SewerCzar:
Will add later:

Converting wights to melee monsters.
Adding aging to all undead.
Upgrading incorporeal undead to ignore AC from armor and weapons.
Making wraiths etc cast fear on movement.

Chilly021:
Only fixing what we can for now. Thanks for the input. Love to see more conversation.

Chilly021:
image

Chilly021:
Only thing I can point to is the current Oak UD implementation. It looks correct. Each of these already have Nightvision. 

  wraith.AddStatus(new NightVisionStatus(wraith));
  wight.AddStatus(new NightVisionStatus(wight));
  spectre.AddStatus(new NightVisionStatus(spectre));
  banshee.AddStatus(new NightVisionStatus(banshee));
  stalker.AddStatus(new NightVisionStatus(stalker));

image

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
```diff
@@ -6,25 +6,41 @@ namespace Kesmai.Server.Game
 {
 	public partial class Banshee : CreatureEntity, IUndead
 	{
+
+		public override CreatureImmunity Immunity { get; set; } = CreatureImmunity.Piercing;
```
sproatba, line 10: pretty sure this won't compile.  I think you need to do public override CreatureImmunity Imunity { get; set; } = CreatureImunity.Piercing || Creature Immunity.Slashing etc.

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
```diff
@@ -6,25 +6,41 @@ namespace Kesmai.Server.Game
...
 		/// <summary>
 		/// Initializes a new instance of the <see cref="Banshee"/> class.
 		/// </summary>
 		public Banshee()
 		{
 			Name = "banshee";
 			Body = 12;
-
 			Alignment = Alignment.Chaotic;
+			DeathResistance = 100;
```
sproatba, line 24: where are these values coming from?  Ice and death have always hurt them, you can hide from them and stun them.  Is there some method behind these values?  I see you quoting dnd references, but this isn't dnd and trying to make it that probably won't work without making it an entirely different game.

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
SewerCzar: image

Swordmaster is a pre-existing entity that Lux made so I think it would actually work.

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
SewerCzar: Other collaborators have determined that base classes of monsters should be D&D manual accurate, changing the balance of almost every zone with undead, so I'm helping out and making undead manual accurate.

I noticed that Death resistance will only give 50% damage reduction, not immunity. I can fix that.

Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Undead/Banshee.cs
sproatba: not sure where you are seeing that code, but it doesn't look like that to me in master.

SewerCzar: I added the other immunities I get what youre saying now I'll fix it so the pull is nice and ready

SewerCzar:
Fixed all the issues pending. Pull looks good now. It's just how undead monsters work.

PR #1720, opened 2022-07-18
2h Fumbles with an offhand 100% of the time

Chilly021:


Leolus (approved):
Only feedback is you can still hold non weapons in left hand while swinging Rhand 2H weapon.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
```diff
@@ -59,9 +59,22 @@ public override int CalculateBlockingBonus(ItemEntity item)
...
-			// Skill Level = 3 =>  ((3 + 1)^2) * 10 = 160 => 1 / 160;
-			// Skill Level = 4 =>  ((4 + 1)^2) * 10 = 250 => 1 / 250;
-			return 1 / (10 * Math.Pow(entity.GetSkillLevel(Skill) + 1, 2));
+			var leftHand = LeftHand;
+			var rightHand = RightHand;
+
+			var leftWeapon = leftHand as IWeapon;
+			var rightWeapon = rightHand as IWeapon;
+
+			if(leftWeapon != null && rightHand != null && rightWeapon.Flags.HasFlag(WeaponFlags.TwoHanded))
```
sproatba, line 68: I think this should be rightWeapon != null since you are only checking for the TwoHanded flag.  could potentially crash if right hand is not a weapon.  Not sure if we can use the Elvis operator, but that is an option as well.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
```diff
@@ -59,9 +59,22 @@ public override int CalculateBlockingBonus(ItemEntity item)
 		/// </summary>
 		public override double CalculateFumbleChance(MobileEntity entity)
 		{
-			// Skill Level = 3 =>  ((3 + 1)^2) * 10 = 160 => 1 / 160;
-			// Skill Level = 4 =>  ((4 + 1)^2) * 10 = 250 => 1 / 250;
-			return 1 / (10 * Math.Pow(entity.GetSkillLevel(Skill) + 1, 2));
+			var leftHand = LeftHand;
```
sproatba, line 62: leftHand and rightHand don't seem to add much to me, can probably ditch them and just use var leftWeapon = LeftHand as IWeapon;  not really a problem though

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
```diff
@@ -59,9 +59,22 @@ public override int CalculateBlockingBonus(ItemEntity item)
 		/// </summary>
 		public override double CalculateFumbleChance(MobileEntity entity)
```
sproatba, line 60: I've been away, but og, and lands don't make 2h necessarily mean 100% fumble like this.  OG, all 2h weapons became 1h weapons eventually, they just got nerfed slightly because of it.  Lands, some 2h weapons were viable 1h and some were true 2h weapons, and it removed the skill portion.  Are we intending to make all 2h true 2h?

sproatba (changes requested):
I think the righthand vs rightWeapon is a crash, rest is just being critical.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
Chilly021: Sounds good, I'll submit a new one. I could do a try catch and bail out. I think that will catch the weapon portion plus adding another conditional. Just makes it so it steps over.

Source/Kesmai.Server/Game/Items/Equipment/Weapons/Weapon.cs
Chilly021: Yes. We can do fumble or absolutely no skill gain from it.

Chilly021: 2 h has base defense by itself

sproatba (approved):

Chilly021:
I'm going to add BBS a specific line of code. I don't want to mess with all 2H categories.

PR #1721, opened 2022-07-19
Weapon Manager and Weapon  DTO

Chilly021:
* Setup a baseline DTO to handle object data for Weapon Properties
* Create a Weapon Manager to set Weapon Properties by Weapon Type and Weapon Level using the DTO

This should be good to go for first round.


Chilly021:
Small Unit test created to test this. 

image

Chilly021:
@jkachhad  @Tenscore  @SewerCzar  - Added a Data Transfer Object to be able to set up the Weapon Manager. Other than Serialization and controlling the text per weapon with enums, there needs to be a quick look over from a Reference standpoint. Due to I was working with half a deck of cards, I had to move this out into a proj file to compile and run unit tests on.

Chilly021:
Additional notes - Weapon Swing , Crit Damage , Crit Hit % and Weapon Flags are Missing for now. 

Weapon flags will be a hard one especially if you want to do customization. (Slashing / Silver / Lawful / BlueGlow / TwoHand)

PR #1722, opened 2022-07-22
Fix Mutators and Naming convention for Kesmai -5

Chilly021:


SewerCzar:
Where is this mutators increasing to veryhigh come from? Please look at other zones for examples. Gem % increases not mutators. Example: Leng, UK, Axe, Oak, Bloodlands. Please check existing content formats before making edits.

image

image

Chilly021:
This was a copy and paste from -4 Kesmai. The only change was to switch the mutator up due to the low gem value. You are asking for people to pick up more gems for less value. I would say change the percentage to be lower, and the mutator to be higher. 

Otherwise you have gem pollution just like UK UD. There have been numerous complaints. That will be the next fix I guess.

Chilly021:
Spoke to Midgit, he will add it. Happy to help.

PR #1738, opened 2022-09-01
kesmai5septmonsters

Miidgiit:
new monsters
new NE region to block wwers
less wraiths
wraiths death spell is 1 less skill
spiders +2 block


Chilly021:
Old PR.

PR #1744, opened 2022-09-07
Blackburrow: Nitpicky stuff - and Add immunities to Blackburrow dragon

Chilly021:


Chilly021:
Covered in another PR.

PR #1762, opened 2022-09-25
Dagger idea

Wirednuke:
Just an idea I'm throwing out there since daggers are "fast" weapons, and it's a way of boosting thief damage a little without making the weapons crazy. 

It's just an idea.. i don't know how this would effect mobs using daggers, etc


sproatba:
Unless it's been fixed, this wouldn't work for normal mobs similar to how we can't make speedy hobs atm.  I think it's a good idea overall to diverge from all 1 round swings.

Chilly021 (approved):

Leolus (approved):

sproatba (approved):

PR #1776, opened 2022-10-15
Update GriffinSkull.cs

Tenscore:
Restoring Mama's skull to her full glory while not gimping the griff.


Chilly021 (approved):

sproatba:
Did we ever manage to fix the stun calculations?  Also worried that this change will really hurt wizards.  The stun on the helmets don't really matter all that much because of how the stun jumps as soon as you equip stun res.  If I remember it's 40 base so 7 vs 10 is negligible.

sproatba (approved):

SewerCzar (approved):

PR #1795, opened 2022-10-27
Creating a Chaotic Rhammer

Tenscore:
Removed Blue-glowing dmg type.
Removed Lawful status. (No Chaotic?)
Serialization.
Description.
--- next.  Getting it to bind to a troll.


Chilly021:
@jkachhad  will need to take a look. I am not aware if this is the right idea.

Tenscore:
I know they need a base class.  Right now they're using the Longbow()

Inheriting from ReturningHammer() Class will bring all the properties. (Lawful, Blue Glowing)

I am aware that properties can be overridden or added, but not sure how to dismiss/take them away.
So I reasoned that it would need its own base class with those modifications.
If we can get it to create it in its sack, and wield it from the sack, it should bind.
If it binds, it should return.

/shrugs
Trying  =)

Chilly021:
No worries, I'll let @jkachhad  take a look. 

Also for disclaimer: 

Just putting this out there. I'm taking myself off of approver until the Halloween event has either expired or the changes have stopped. This is just to make sure we are not causing any fun weirdness here.

You may reach out to @jkachhad otherwise, if you need quicker feedback.

sproatba (approved):

Chilly021 (approved):

zrikzlok (dismissed):

Chilly021 (approved):

Leolus (approved):

zrikzlok (approved):

PR #1803, opened 2022-12-06
Move Ram Staff and Unholy Scepter to Wield

Chilly021:
Currently this is the only way we have found out to make sure people don't carry a corpse with the item on it. This ensures the player has to attempt or at least traverse to the region to get the item. (similar to OL staff)

Leng: Killer Rabbit 
Bloodlands: The Queen

### DIFF
--- a/Segments/Leng.mapproj
+++ b/Segments/Leng.mapproj
@@ -79115,6 +79115,8 @@
     
     critter.CombatantChangeInterval = TimeSpan.FromSeconds(10.0);
     
+    critter.Wield(new RamStaff());
+    
     critter.Attacks = new CreatureAttackCollection()
     {
         {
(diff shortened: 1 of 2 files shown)


sproatba:
With the logic limiting item use, can they actually wield these items?

Chilly021:
Good question, I would assume CreatureEntity is a little different when determining attack, but they both inherit from MobileEntity. 

Carfel has wield (both daggers, one neutral, one does not have alignment), but no limiting feature. 
Sword Master has wield, but no limiting feature. 
DoomOrc has wield, but no limiting feature. 

I have absolutely no way to test it without a functional test (after deploy).

sproatba (approved):

zrikzlok (approved):

PR #1815, opened 2022-12-30
Adjusting Mama to be able to fly/hit 3x away. DB Ice is super infrequent currently. 

Chilly021:
1) Mama has always been able to attack 3 hexes away in other emulators. This is a fix since Charge only attacks 2x away. 

2) Due to CreatureSpellCollection(35.0) is at 35% spell to melee ratio and also a timespan is being applied for 15 seconds, the icestorm is far less frequent than any other version. Just to slightly affect db ice percentage bumping the rate up by 5 seconds.

### DIFF
--- a/Segments/Axe Glacier.mapproj
+++ b/Segments/Axe Glacier.mapproj
@@ -107034,7 +107034,7 @@
         Experience = 213980,
         
         CanFly = true,
-        CanCharge = true,
+        CanJumpkick = true,
         CanStrikeCritically = false,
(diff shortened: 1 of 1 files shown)


Leolus (approved):
Just a note that part of the reason the ice storm was set so high is icestorm is actually a dps drop for her overall. It's much easier to solo her at a lower level the more frequently she icestorms. Not opposed to the change just giving some context.

zrikzlok (approved):

sproatba (approved):

PR #1846, opened 2023-02-08
Fix Gaunts for Bloodlands

Chilly021:


Chilly021:
@jkachhad when you have a second.

PR #1853, opened 2023-02-15
MinorBBupdate

Miidgiit:
fixed some inaccessible areas.
fixed grates
swapped some BODY IDs
edited -200 access to the South Mines.
Upped HP on minor bosses
Upped Undead's poison from 3 to 6.
More mobs in -300
Changed some poison tiles in the PIT

### DIFF
--- a/Segments/Blackburrow.mapproj
+++ b/Segments/Blackburrow.mapproj
@@ -8799,26 +8799,20 @@
         </component>
       </tile>
       <tile x="43" y="-12">
-        <component type="WallComponent">
-          <wall>333</wall>
-          <destroyed>342</destroyed>
-          <ruins>44</ruins>
-          <indestructible>true</indestructible>
-        </component>
         <component type="WaterComponent">
-          <ground>22</ground>
+          <ground>623</ground>
           <movementCost>3</movementCost>
         </component>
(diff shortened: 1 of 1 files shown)


Segments/Blackburrow.mapproj
```diff
@@ -91494,7 +91559,7 @@ if (spokeRecently(source))
 {
     Name = "Potion Master Nook",
     Alignment = Alignment.Chaotic,
-    Body = 388,
+    Body = 387,
```
Chilly021, line 91562: Can we confirm that this is compiled and can be used?

Chilly021 (commented):
Can we confirm that this is compiled and can be used?

Segments/Blackburrow.mapproj
Miidgiit: I would assume yes.. while they dont match with the Bestiary. 883-388 work but dont match the website. 388 is suppose to be a shaman dogman. and 387 is a variant shaman dogman.. but 388 is a wolf with a sword.

Chilly021 (approved):

zrikzlok (approved):

sproatba (approved):

PR #1858, opened 2023-02-21
CGS Fix hopefully

Chilly021:
CGS Fix hopefully


Chilly021:
One sec - need to re-evaluate the Code for Konran/Silk. Seems the issue is further up the chain and need to pull in if the Killing blow was landed / if both items were bound to the character.

PR #1873, opened 2023-02-28
Shu

Miidgiit:
New area in -200
New boss who uses whirlwind
2 new minibosses
MU only area with teleport phase puzzle.
gnoll raider can swim now.


Chilly021 (changes requested):
I'm seeing a lot of overlapping components here. You might just remove statics and keep your stair case. 
image

Chilly021:
Some of the water teleports go no where. Is that intentional? 
image

Chilly021 (commented):
You can remove these fellas. These were just a safeguard. Now that we have the proper "AddStatus", no longer need these two. 

tiger.AddImmunity(typeof(BlindSpell));
tiger.AddImmunity(typeof(FearSpell));

Chilly021 (commented):
Confirmed with @Miidgiit  - we are okay here. Strictly consmetic

Chilly021 (approved):
It was strictly cosmetic.

Leolus (approved):

zrikzlok (approved):

PR #1881, opened 2023-03-01
Make Doombow great again

Wirednuke:
I've been playing with the doom bow... it's weak for the amount of risk and time to use it. Propose doubling the damage to see what it feels like?


Tenscore:
I approve this message.   ;p

Chilly021 (commented):
The issue is for this, we probably want to pull this down into the segment of Oakvael versus adding changes directly here. The reason is, if something is extremely off and it needs testing, server side compiles take a bit here. 

I can help you with that, if you want.

sproatba (approved):

Chilly021:
Closing due to Conflicts. We can re-address in segment or in code.

PR #1893, opened 2023-03-11
Fix Language for Puck and Update Enu

Chilly021:
* Tell users to go to Yuri if they haven't started the Quest
* Add actual text for missing Enu Strings


sproatba (approved):

Chilly021:
Thank you!

Chilly021 (commented):
@yarrim  Good Catches and this is great feedback! Per PR commit: 

Fixing per Kora's Feedback
* Yep I removed the null check on Quest I added originally as a mistake. It never gets hit though it seems, but regardless good to have. Thank you!

* Good catch on the khopesh. I have had to google the spelling like 4 times and yet, still didn't learn my lesson here again. English / Egyptian translation is hard.

* Corrected the Yuri verbiage to replace colon with comma.

yarrim (approved):

sproatba (approved):

Leolus (approved):

PR #1896, opened 2023-03-14
Example Code HolyWeapon

Chilly021:
Check to see if mobile entity Type is IUndead then CalcDamage with weapon.

This is not for me, just trying to be helpful.

image


Chilly021:
Can always pull this back up for reference. Closing for now.

PR #1908, opened 2023-03-16
Bloodlands weapons drop at feet

Chilly021:


Segments/Bloodlands.mapproj
```diff
@@ -225237,17 +225237,21 @@
...
 	
     return fighter;
 ]]></block>
         <block><![CDATA[]]></block>
       </script>
-      <script name="OnDeath" enabled="false">
+      <script name="OnDeath" enabled="true">
         <block><![CDATA[]]></block>
         <block><![CDATA[
+    if (killer is PlayerEntity player)
```
Leolus, line 225250: Does it count as killed by player if a hex damage based spell is cast? May want to test weird death scenarios once this is in if relevant.

Chilly021: So the test was originally done on hummer boss for guardian. The issue there is he can only be killed by blue glow weapons at the time. Can try to kill from afar but I think if the crit moves at the same time, it will put it at player feet.

PR #1926, opened 2023-03-25
Both Rhammers drop on ground (similar to -4 Kes Rhammer) - Adventurers need to Adventure

Chilly021:
Giant
Since it is 100% - just drop at feet. No more corpse teleportation.

Ydnac
Add D3 roll for the removal of the group loot. (33.33%) Evaluate if lands on 3 and drop at feet, otherwise bail out. No more corpse teleportation.


yarrim:
Can non-player Entities SayTo() each other? Not sure how often that will happen, or even if it could in this case, but I would restructure so that the SayTo()s are inside the killer is PlayerEntity branch just in case.

zrikzlok (approved):

Leolus (approved):

sproatba (approved):

PR #1951, opened 2023-04-11
BB New Area n Mob updates

Miidgiit:
Kesmai: -5 Bloodlands hex isnt water anymore
Minibosses have Fear Resist.
Undead Dragon: DBL is buffed. Poison Cloud Buffed slightly Changes in mob density in -200 and -300.
New area added in -300. no mobs
New +100 area added. no mobs.


Chilly021:
This is borked. We are re-submitting.

PR #1956, opened 2023-04-13
Hall of Legends: Fix Shopkeeper

gnummers:
Fix so that shopkeeper counter can be accessed correctly from the north


gnummers:
Sorry, going to resubmit correctly.

PR #1963, opened 2023-04-17
Leng Updates - Griff Cliff / Kosh Gaunts

Chilly021:
* Split Spawns Up - Keeping Relatively the same number. (N / W / E )

FROM: 
image
TO: 
image

* Kosh Gaunts drop on Ground
* ITreasure on Ram Staff


zrikzlok (dismissed):

Chilly021 (commented):
Resubmitting without the following change later today: 

image

PR #1970, opened 2023-04-20
ArcherTune

Tenscore:
Per request:
Dropped Archer pen down a notch in Axe, Oak, Leng, and UK.
Leaving BL, and BB for their makers as they're still a WiP.

### DIFF
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -113956,6 +113956,7 @@
     {
         MaxHealth = 80, Health = 80,
         BaseDodge = 16,
+        BasePenetration = ShieldPenetration.None,
 
         Experience = 2902,
 
--- a/Segments/Oakvael.mapproj
+++ b/Segments/Oakvael.mapproj
@@ -111398,7 +111398,6 @@
         BaseDodge = 14,
         HideDetection = 21,
         Experience = 1507,
-        BasePenetration = ShieldPenetration.VeryLight,
     };
     
     goblin.Attacks = new CreatureAttackCollection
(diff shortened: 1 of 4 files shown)


Segments/Oakvael.mapproj
```diff
@@ -111552,7 +111551,6 @@
         BaseDodge = 14,
         HideDetection = 21,
         Experience = 1755,
-        BasePenetration = ShieldPenetration.VeryLight,
```
Chilly021, line 111555: For Very light removal, you might want to switch that to "none".

Segments/Oakvael.mapproj
```diff
@@ -111398,7 +111398,6 @@
         BaseDodge = 14,
         HideDetection = 21,
         Experience = 1507,
-        BasePenetration = ShieldPenetration.VeryLight,
```
Chilly021, line 111401: For Very light removal, you might want to switch that to none. Otherwise it inherits from weapon.

Chilly021 (commented):
Few comments around ShieldPenetration.Verylight. By removing this it will just take into account whatever the weapon has at the time. You can use None. 

BasePenetration = ShieldPenetration.None

[WorldForge]
	public enum ShieldPenetration
	{
		None = 0,

		VeryLight	= 1,
		Light		= 2,
		Medium		= 3,
		Heavy		= 4,
		VeryHeavy	= 5,
		Full		= 6,
	}

Segments/Oakvael.mapproj
Tenscore: Just noticed Crossbow.cs would set it back to Light.

Nice catch, thx!

Segments/Oakvael.mapproj
Tenscore: Got it.

Tenscore:
Seems it may be best to attack this at Crossbow.cs, and Longbow.cs

Leolus (approved):

zrikzlok (approved):

PR #1971, opened 2023-04-21
Dartanian's robe/jacket

Wirednuke:
Here is a new item to drop from Dartanian in BB. 

Two issues:

1) I don't know the correct spell inscription ID so it'll probably come up with the wrong spell ID until I find out what it is

2) I'm not sure if the CanEquip function will work for a 'robe' inside segment code.


Chilly021 (approved):

Chilly021 (commented):
All for testing - but again - the spell inscription ID needs to be corrected more than likely

Wirednuke:
> All for testing - but again - the spell inscription ID needs to be corrected more than likely

Yea I agree, it's just been a week and Lux hasn't responded and I'm tired of waiting lol.

sproatba (approved):

Leolus (approved):

PR #1982, opened 2023-04-26
Update Spell.cs

Tenscore:
should be around 5minutes.
Shield spell under x2 is 120seconds.
No way to be sure but to test!


Tenscore:
image

Leolus (approved):

zrikzlok (approved):

PR #1984, opened 2023-04-26
Javelin Introduction

Wirednuke:
The weapon itself is pretty straight forward. It's a returning Javelin, I want it to do piddly damage in melee and up close and devastating damage when thrown off hex. 

The onHit function needs some feedback. I think it will work but there's probably a cleaner way of doing this that I'm not familiar with? @jkachhad


Wirednuke:
I'd also like it to be main hand only (foresee it as spear and shield), not sure how to accomplish that.

Wirednuke:
> The OnHit function could be simplified to something like
> 
> 
> 
> `attacker.CalculateThrowDamage *= (source.GetDistanceToMax(entity) + 1)`
> 
> 
> 
> That'll make it do more damage the further it's thrown.
> 
> 
> 
> `attacker.CalculateThrowDamage *= source.GetDistanceToMax(entity) ? 3 : 1`
> 
> 
> 
> That'll keep current logic. 

I like it, thanks for the help!

Segments/Blackburrow.mapproj
```diff
@@ -796,6 +796,92 @@
...
+                case 1:
+                    {
+                        break;
+                    }
+            }
+        }
+        
+        public override void OnHit(MobileEntity attacker, MobileEntity defender)
+        {
+            attacker.CalculateThrowDamage *= (source.GetDistanceToMax(entity) + 1);
```
Chilly021, line 880: attacker.CalculateThrowDamage *= (attacker.GetDistanceToMax(defender) + 1);

Chilly021 (commented):
Source / Target are Attacker / Defender in this scenario.

Segments/Blackburrow.mapproj
Wirednuke: Ok, so I need to specify that it's defender there?

Wirednuke:
Made changes for @Chilly021, and copy/modified the code from BBS/BL Sword to make the javelin right hand only.

Segments/Blackburrow.mapproj
```diff
@@ -796,6 +796,105 @@
...
+                case 1:
+                    {
+                        break;
+                    }
+            }
+        }
+        
+        public override void OnHit(MobileEntity attacker, MobileEntity defender)
+        {
+            attacker.CalculateThrowDamage *= (source.GetDistanceToMax(defender) + 1);
```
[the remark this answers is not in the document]
Wirednuke: Missed it, my bad. All fixed now.

Chilly021 (approved):

Chilly021:
Need to put this in draft until Lux gets back from his trip.

Leolus (approved):

zrikzlok (approved):

PR #1996, opened 2023-05-05
Undoing Javelin Changes

Wirednuke:
Reverting the code, will revisit this after I talk with @jkachhad. Probably end up going the ApplyDamage route.


Chilly021 (approved):

Leolus (approved):

sproatba (approved):

Leolus:
Looks like the distance issue is it needs to reference a location not an entity. See the implementation here: https://github.com/jkachhad/Stormhalter/blob/c595eae3b8f145b2f7b2a7e5db21dc0c2162943d/Source/Kesmai.Server/Game/Entities/Mobiles/Creatures/Dragonkin/IceDragon.cs

PR #2000, opened 2023-05-07
Librarian boss and Emon update

Miidgiit:
Emondeus= Demon now

Soul Sucker has penetration= goal is to always hit.. always drain your life..but for very little.. But little does add up.

Added the Librarian boss to -1  its a skeleton with a flail. I've heard talks of it being from OG so I thought I'd pop it out there.

most -1 mobs are 8 hp. It's 40+ hp with 1 skill higher chance to hit. 1 hrish respawn.


Segments/Kesmai.mapproj
```diff
@@ -145233,18 +145233,18 @@ else
         
         IsTethered = true,
     };
-    dragon.AddStatus(new NightVisionStatus(dragon));
+    demon.AddStatus(new NightVisionStatus(dragon));
```
Chilly021, line 145236: You need to change the new nvstatus(dragon) to new nvstatus(demon), otherwise compiler issues.

Chilly021 (changes requested):
One comment. The constructor needs to pass in the demon param. I'll be back tonight if you need anything else.

Chilly021 (commented):
Does the librarian need a unique body type? Or is a default skeleton all you wanted?

Miidgiit:
simple skeleton for now.

Miidgiit:
i changed his color slightly. (end of my changes)

Chilly021 (approved):

Leolus (approved):

sproatba:
Looking at this on my phone so I don't have the editor to help, but it looks like a kobold with all zeros was added.  Not sure what that will do.

Miidgiit:
image

So yeah. I have no clue how to actually delete a made region.. I dont see an option anywhere so I just turned it to zero.

zrikzlok (approved):

PR #2004, opened 2023-05-11
Jacket Tweak

Wirednuke:
See Code


yarrim (approved):
We can't see the WorldEntity code, so I can't be 100% sure this is crashproof, but getSkillLevel() isn't defined at the PlayerEntity class, so it should exist at MobileEntity.
I would probably keep the tests separate, just in case. Test that the entity is a player and a thief, then get its skill level and test that. If this code gets reused as a template for a criteria that IS player-only, it will break.

Segments/Blackburrow.mapproj
```diff
@@ -787,7 +787,10 @@
         
         public override bool CanEquip(MobileEntity entity)
         {
-            if (entity is PlayerEntity { Profession: Thief })
+        
+            var skilllevel = entity.GetSkillLevel(Skill.Magic);
+            
+            if (entity is PlayerEntity { Profession: Thief } & skilllevel >= 15)
```
Chilly021, line 793: `I would split these conditionals. Just from clarity standpoint. 

var skilllevel = entity.GetSkillLevel(Skill.Magic);
var isUsable = false;

/* 
Making sure the character is thief and level 15
*/
 if (entity is PlayerEntity player)
            {
                if (player.Profession == Profession.Thief && skilllevel >= 15)
                {
                    isUsable = true;
                }
            }
        }
        
        return isUsable;
               
`

Wirednuke:
Made recommended changes.

yarrim (approved):

sproatba (approved):

PR #2015, opened 2023-05-14
BB-100

Miidgiit:
New area to the North East and South.
added a few new regions.

PIT: turned the bridges SOLID near NOOK.

Plaguesayer: up its EXP. Lowered its SKILL by 2.


Chilly021 (commented):
It looks like you have a lot of compilation errors. 

image

PR #2036, opened 2023-05-19
BB - Wizard Lair -- Fix Boss loot drop and other tweaks

Wirednuke:
Making a detailed list of my changes here to hopefully make review easier.... Made the following changes:

- Removed loot pack entries from the Spirit of Al'Kabor in the onDeath() portion that summons him, and added items to his backpack via the creature class in segment code. 
- Set a static location for the 'Spirit of Al'Kabor' to spawn on the death of 'rotting mummy' instead of having it spawn on killer.location
- Switched the art on the 'Spirit of Al'Kabor' until the new player created art is fixed to allow proper targeting
- Wrote a method (getRandomQuality) that randomly generates an ItemQuality and passes it to loot in the backpack
- Added a boolean expression to the DeathLightningSpell that should hopefully prevent it from spamming lightning continuously 
- Substantially increased the population of Melee mummies based on player feedback


Segments/Blackburrow.mapproj
```diff
@@ -1298,13 +1298,66 @@
...
+    ItemQuality quality = ItemQuality.Common;
+
+    if (chance < 500)
+    {
+        return quality;
+    }
+
+    if (chance >= 500 && chance < 750)
+    {
+        quality = ItemQuality.Uncommon;
```
Chilly021, line 1314: You probably want to set this to common and the local var to uncommon. Change < 500 being uncommon.

Chilly021 (commented):
You probably want to set this to common and the local var to uncommon. Change < 500 being uncommon.

Chilly021 (approved):

Leolus (approved):

zrikzlok (approved):

PR #2059, opened 2023-05-28
Make the traps move after they cast

Wirednuke:
Just experimenting with making the wizard lair traps move after they cast. 

Using OnSpellTarget to start a kill timer for 1 minute. Then the location spawner should take care of the rest. 

Only question: Can invulnerable traps be killed by the Kill command?

### DIFF
--- a/Content/Localization/Enu.xml
+++ b/Content/Localization/Enu.xml
@@ -289,9 +289,9 @@
   <entry index="6000102">idol</entry>
   <entry index="6000103">warhammer</entry>
   <entry index="6000104">Hat</entry>
-  <entry index="6000104">vial</entry>
   <entry index="6000105">cuirass</entry>
   <entry index="6000106">gauche</entry>
+  <entry index="6000107">vial</entry>
   <entry index="6005000">corpse</entry>
 
   <entry index="6100000">{0}</entry>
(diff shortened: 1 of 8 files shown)


Wirednuke:
Hold off on reviewing, more changes coming and I can't mark as draft on the app

Wirednuke:
Ok we are good now... ready for review:

- Added code to make traps check their location when they spawn for another trap and respawn if they do. 
- After traps cast they will kill themselves and respawn at a different location
- Changed traps from IsInvisible to IsHiding and added Level's hoping that they are hidden but able to detect each other. 
- Added a mechanism in OnIncomingPlayer method that unhides the trap when a sufficiently leveled thief is in proximity.

Chilly021 (approved):

zrikzlok (dismissed):

sproatba:
Seems to me just having the trap kill itself and giving it a region to spawn in would make it move,  making it teleport around and whatnot seems overly complicated to me.

sproatba (approved):

Wirednuke:
Hold on... don't merge.. conflicts now for some reason

Chilly021 (approved):

zrikzlok (approved):

sproatba (approved):

PR #2068, opened 2023-06-03
EmondeusTweak

Miidgiit:
spell swap with Emondeus.
DragonBReath is now FireStorm

+ a bunch of other stuff I didnt add myself...


Chilly021 (dismissed):

Chilly021 (approved):

Chilly021 (commented):
We should be good on the auto added enables. It happened for me in bloodlands too via world forge save. Regardless it's just adding the xml attribute for enable to true, which by default it is already. 

The true change is just using fireball. I can help you create an entity when I get freed up. Similar to all the trap Entity / dark efreets I've created you just can program the entity with any sounds/behavior you want without relying on inherited values of another class.

Leolus (approved):

zrikzlok (approved):

PR #2085, opened 2023-06-15
Konran Logic Update

Chilly021:
* Bail out sooner if Konran has already given a silk finery. No need to execute the rest.


yarrim (approved):
I think this is correct. I'm not familiar enough with Konran to know if he wanders around after the turn-in. if so, he might spam the Director with each step.

Leolus (approved):

yarrim (approved):

PR #2098, opened 2023-07-07
Test of HealerAI

Wirednuke:
Created a HealerAI that is a modified CombatAI that should make the NPC heal its comrades and itself when they are hurt. 

I created a new entity from the gnoll entrance casters and populated a few of them. 

It compiles with zero errors.


Chilly021:
This might take a little more than a while to code review: 

image

Chilly021 (commented):
As mentioned, a lot of different changes here. I can't read them in the code edits or the xml compare tool I have. 

image

PR #2112, opened 2023-07-17
Add mageGlove prototype

Wirednuke:
Added Mage gauntlets. Only wearable by Wizard and Thaum. They will have good attack rating, but low physical damage (less than Kosh Gaunts in terms of physical damage). 

If you are a wizard: They should proc magic missile (3x Magic Level in damage) on a successful hit. 

If you are a Thaum: They will take the damage done and split it as a heal to your summons. 

They will drop off Wizard Al'Kabor at a 50% rate. 

Will need to switch the artwork when Tenser's mageGloves get compiled in. 

Future ideas:
Add a quest to turn them into stat gloves for non-hand users. 
Add Quality
Adjust damage/heal rate as needed. 

For now, I just want to make sure they work. 

Compiled with zero errors.


Chilly021 (approved):

Wirednuke:
Applying a lesson I learned from my cleave Pike.

Chilly021 (approved):

zrikzlok (approved):

Leolus (approved):

PR #2114, opened 2023-07-18
Fix Cleave not giving XP

Wirednuke:
I think I figured out why Cleave wasn't giving XP. Need to register the outgoing damage to the character that is applying it.


Chilly021 (commented):
As Lux mentioned, add RegisterIncomingDamage as well.

Chilly021 (approved):

zrikzlok (approved):

Leolus (approved):

PR #2118, opened 2023-07-21
Troubleshooting Gaunts on Thaums

Wirednuke:
Gaunts are not doing anything for Thaums, I suspect the LINQ query isn't picking up anything. 

Changed LINQ query from Directors = players to all friendlies to check if my list was null


Chilly021 (commented):
Still on cell phone, but I think you can use the summons from IPhantasm or djinn to search by? Player group seems odd here. I'll need to look when I get home late tonight

Wirednuke:
> Still on cell phone, but I think you can use the summons from IPhantasm or djinn to search by? Player group seems odd here. I'll need to look when I get home late tonight

The original method I used pulled CreatureEntity's and then looked for entities with Director = player. That didn't seem to work, so I just figured I'd go easy and grab all the players visible.

Chilly021 (approved):

Leolus (approved):

zrikzlok (approved):

PR #2137, opened 2023-08-02
Amulet quest and messaging changes

Wirednuke:
Things changed:

- Added dialog to vendor to make the amulet leveling process less confusing. 
- changed amulet messaging to show leveling progress when amulet is right clicked. 
- Made amulet bind, and it can now only be equipped by spell casters. 
- Changed the leveling curve so XP required now doubles correctly every level. 
- added FrogGauntlets and BL caster/thief weapons to the list of weapons that will level the amulet (yay thieves)
- raised Thaum Magegauntlet self-heal percentage
- Thaum gauntlets will now only heal other players if the target is a monster.


Wirednuke:
The Brick virus is spreading to Bloodlands.

Chilly021 (approved):

sproatba (approved):

PR #2169, opened 2023-09-02
Update for Pearl Usage

Chilly021:
Update for Pearl Usage


Chilly021:
@Miidgiit  got with Lux on this and we are all fixed up. Brick will probably remove the Send command.

PR #2246, opened 2025-03-18
Add poison multiplier to base dagger

Wirednuke:
Added a property to dagger base class that allows venom to be scaled across all daggers. Also implemented the multiplier by overriding the base OnHit method to check for poison on the dagger and modifier the potency with the multiplier if it's greater than 1. 

I put a setter on the property so we could make other wearable items that could have multipliers that further adjust the potency.


Wirednuke:
Just to be clear, I have no intention of modifying the base multipliers in any of the OG weapons. This is just to make it easier to scale weapons in new content.

Miidgiit (approved):

zrikzlok (approved):

n099y (approved):

PR #2259, opened 2025-08-02
Add OR (|) and NOT (!) operators to targeting filters

koorbmeh:
- Add support for | (OR) operator to combine multiple filter conditions
- Add support for ! (NOT) operator to negate filter conditions
- Implement FilterCondition class to handle complex filter parsing
- Add ParseFilterConditions method to split and parse filter strings
- Add ApplyFilterCondition method to handle individual filter logic
- Maintain backward compatibility with existing filter syntax
- Support both basic filters (hostile, near, distant) and advanced filters (distance(), serial())
- Examples: [distant | near], [!hostile], [!hostile | !npc], [poisoned | feared]

### DIFF
--- a/Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.Visibility.cs
+++ b/Source/Kesmai.Server/Game/Entities/Mobiles/MobileEntity.Visibility.cs
@@ -80,6 +80,89 @@ public abstract partial class MobileEntity
 	};
 
 	/// <summary>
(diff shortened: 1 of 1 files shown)


n099y:
This should be AST (Abstract Syntax Tree) so it allows endless grouping and future proofed. ` @[ (hostile:casting) | (!distant) ] `

n099y (approved):
Looks good, thank you for this.

Chilly021 (approved):

PR #2263, opened 2025-08-13
Improve filter system performance and structure

koorbmeh:
- Replaced FilterNode hierarchy with cleaner FilterExpression classes
- Added high-performance FilterState with boolean arrays to eliminate list copying (which was still happening even though it looked like you tried to eliminate it)
- Optimized filter application logic for better performance
- Fixed healthy filter logic
- Improved near filter logic
- Improved OR operation handling with priority-based evaluation
- Maintain backward compatibility while improving internal implementation


n099y:
Closed due to age and new conflicts

